PR #1, opened 2021-06-15
Port to CMake

PureTryOut:
This project uses KDE's KCalendarCore. However, that package can only be
used with build systems that support CMake dependencies, which QMake is
not one of. KCalendarCore probably shipped a pkgconfig file in the past,
but this has not been the case for a quite a while as far as I can see.

Since using QMake thus prevents using a current KCalendarCore, switch
the build system to CMake instead.

This is a recreation of https://git.sailfishos.org/mer-core/nemo-qml-plugin-calendar/merge_requests/82 as sadly you guys moved away from the FOSS Gitlab.


neochapay:
https://github.com/KDE/kcalendarcore/commit/e497e0a4cdb3a4ea7eef42e3d5ace9068c1b66e0 fixed in kde ^_^

PureTryOut:
@neochapay nice but qmake is still being deprecated by Qt in favor of CMake (they don't even use it themselves anymore from Qt6 and onwards), so switching to CMake should still be done eventually.

CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+set(CMAKE_AUTOUIC ON)
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+include(FindPkgConfig)
+include(FeatureSummary)
+include(GNUInstallDirs)
+include(CTest)
+
+set(INSTALL_QML_IMPORT_DIR "${CMAKE_INSTALL_FULL_LIBDIR}/qt/qml"
```
dcaliste, line 16: At the moment, SailfishOS is using `$LIBDIR/qt5/qml` as the QML import dir. If the `qt5` part is an issue, it should be set as a variable.

tests/tst_calendarevent/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+	${PROJECT_SOURCE_DIR}/src
+	${TARGET_MOC_DIR})
+
+target_link_libraries(tst_calendarevent
+	Qt5::Qml
+	Qt5::Test
+	KF5::CalendarCore
+	nemocalendar)
+
+add_test(tst_calendarevent tst_calendarevent)
```
dcaliste, line 15: Like in `sailfishos/mkcal`, tests should be installed, for instance with:

```
if(INSTALL_TESTS)
	install(TARGETS tst_calendarevent
		DESTINATION /opt/tests/nemo-qml-plugins-qt5/calendar)
endif()
```

see old `common.pri`.

dcaliste (changes requested):
The rpm/nemo-qml-plugin-calendar-qt5.spec file should be updated also to use CMake instead of qmake., something like:
```
--- a/rpm/nemo-qml-plugin-calendar-qt5.spec
+++ b/rpm/nemo-qml-plugin-calendar-qt5.spec
@@ -6,6 +6,8 @@ Release:    1
 License:    BSD
 URL:        https://github.com/sailfishos/nemo-qml-plugin-calendar
 Source0:    %{name}-%{version}.tar.bz2
+BuildRequires:  cmake
+BuildRequires:  extra-cmake-modules >= 5.75.0
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5Gui)
 BuildRequires:  pkgconfig(Qt5Qml)
@@ -46,14 +48,12 @@ BuildRequires:  pkgconfig(KF5CalendarCore)
 %setup -q -n %{name}-%{version}
 
 %build
-
-%qmake5 
-
+%cmake -DBUILD_TESTING=ON
 make %{?_smp_mflags}
 
 %install
 rm -rf %{buildroot}
-%qmake_install
+%make_install
 
 %files
 %defattr(-,root,root,-)
```

tests/tst_calendarevent/CMakeLists.txt
PureTryOut: I never understood this, why exactly? I have no problem adding it under a conditional block (so at least I have the option to not install them) but I'd like to understand why you install them in the first place.

dcaliste: Sorry, I'm not working at Jolla. I'm part of the community also. Maybe @pvuorela could answer.

I guess that they want RPMs to easily deploy tests on a test infrastructure. Now, this does not necessary imply to put the installation procedure in the build system itself of the project. It could be done at the spec level for specific SailfishOS builds. I agree. But well, I guess it's historical and maybe easier also than writing `install -m 644 -p...` multiple times in the spec file.

pvuorela: We do automatic nightly run for the unit tests and the setup requires that there's the tests.xml and the test binaries installed.

PR #2, opened 2021-06-16
Add a list model for a bunch of events

dcaliste:
This is adding a new list model, using the same roles as an agenda model but used when we have a bunch of events based on their instanceIdentifiers.

See mkcal!60 (merged) for more details.

@chriadam and @pvuorela , this is a copy of the original MR from Gitlab, see https://git.sailfishos.org/mer-core/nemo-qml-plugin-calendar/merge_requests/84


chriadam:
The motivation for this PR is to allow showing e.g. succeeded and / or failed events in the sync log UI, right?  If so, can you please add "Contributes to JB#49486" to the commit messages?

chriadam:
@pvuorela @llewelld @blammit any reasons not to merge this one?

dcaliste:
> The motivation for this PR is to allow showing e.g. succeeded and / or failed events in the sync log UI, right?

Right, this provides a list model created from a list of instance identifiers. So if you have a list of events that failed to sync properly, you can display them using the UI code displaying events in the month view for instance, since the added list model is exposing the exact same roles than the agenda model.

I've just added the corresponding JB.

PR #3, opened 2021-06-23
[nemo-qml-plugin-calendar] Use new rawExpandedIncidences() to fetch a…

dcaliste:
…ll occurrences.

After sailfishos/mkcal#2 will be accepted, to avoid the deprecation warning, it's better to use the new `rawExpandedIncidences()` API. This is the first commit.

But I've added a second commit actually, that is removing the use of `rawExpandedIncidences()` completely in favor of the KCalendarCore `OccurrenceIterator` instead. I think this is cleaner and allow to join the two for loops of the previous implementation. This makes actually the `mKCal::ExtendedCalendar::rawExpandedIncidences()` useless now, because as far as I know it was the only place where this method was used. We could even remove it completely from mKCal since it's just a wrapper around the `OccurrenceIterator`, making one less `mKCal::ExtendedCalendar` specific function compared to upstream.

@chriadam and @pvuuorela, what do you think ? Should we keep only the first commit and discard the second or keep the second also ? If we keep the second and decide to remove the `rawExpandedIncidences()` from mKCal, we will loose the extensive tests that are done there to ensure the expansion is working for corner cases like local time, all day events... Maybe I can move them instead ?

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -49,6 +49,7 @@
 #include <KCalendarCore/Recurrence>
 #include <KCalendarCore/RecurrenceRule>
 #include <KCalendarCore/MemoryCalendar>
+#include <KCalendarCore/OccurrenceIterator>
 
 // libaccounts-qt
 #include <Accounts/Manager>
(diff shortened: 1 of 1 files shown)


src/calendarworker.cpp
```diff
@@ -715,30 +716,28 @@ void CalendarWorker::setNotebookColor(const QString &notebookUid, const QString
...
-        filtered.insert(occurrence.getId(), occurrence);
+    foreach (CalendarData::Range range, ranges) {
+        KCalendarCore::OccurrenceIterator it(*mCalendar, QDateTime(range.first.addDays(-1)),
+                                             QDateTime(range.second.addDays(1)).addSecs(-1));
+        while (it.hasNext()) {
+            it.next();
+            if (mCalendar->isVisible(it.incidence()) &&
+                !excluded.contains(mCalendar->notebook(it.incidence()))) {
+                const QDateTime sdt = it.occurrenceStartDate();
+                const KCalendarCore::Duration enlapsed
```
pvuorela, line 729: "Enlapsed"?

src/calendarworker.cpp
```diff
@@ -715,30 +716,28 @@ void CalendarWorker::setNotebookColor(const QString &notebookUid, const QString
...
-        occurrence.startTime = exp.first.dtStart;
-        occurrence.endTime = exp.first.dtEnd;
-        filtered.insert(occurrence.getId(), occurrence);
+    foreach (CalendarData::Range range, ranges) {
+        KCalendarCore::OccurrenceIterator it(*mCalendar, QDateTime(range.first.addDays(-1)),
+                                             QDateTime(range.second.addDays(1)).addSecs(-1));
+        while (it.hasNext()) {
+            it.next();
+            if (mCalendar->isVisible(it.incidence()) &&
+                !excluded.contains(mCalendar->notebook(it.incidence()))) {
```
pvuorela, line 727: Should this filter out non-event type incidences? 

Cosmetic detail: operators at the beginning of the line help to separate if statement from the following code too.

src/calendarworker.cpp
dcaliste: Sorry, I meaned "elapsed".

src/calendarworker.cpp
dcaliste: You're right, I forgot to filter out non events. I've added a check.

src/calendarworker.cpp
```diff
@@ -715,30 +716,29 @@ void CalendarWorker::setNotebookColor(const QString &notebookUid, const QString
...
+                const QDateTime sdt = it.occurrenceStartDate();
+                const KCalendarCore::Duration elapsed
+                    (it.incidence()->dateTime(KCalendarCore::Incidence::RoleDisplayStart),
+                     it.incidence()->dateTime(KCalendarCore::Incidence::RoleDisplayEnd),
+                     KCalendarCore::Duration::Seconds);
+                CalendarData::EventOccurrence occurrence;
+                occurrence.eventUid = it.incidence()->uid();
+                occurrence.recurrenceId = it.incidence()->recurrenceId();
+                occurrence.startTime = sdt;
+                occurrence.endTime = elapsed.end(sdt);
```
pvuorela, line 738: If there's a DST change in the middle, the endTime is not start+duration? Would that be a problem here?

dcaliste: That's a good question. Here is a snippet from the old code from `mKCal`:
```
                    const QDateTime tsStartTime = kdatetimeAsTimeSpec(timeInInterval, tz);
                    const QDateTime tsEndTime = Duration(startTime, endTime).end(tsStartTime);
```
So at least, I'm not introducing any regression here.

That being said, could we do better ? Here are two situations that may arise:
1. we have a recurring event to bring our bat pet out from 1am to 3am, the bat requires 2 hours to feed.
2. we have a recurring event to work as a night guard from 22pm to 6am the next day.

In case of the first event, we're interested in the duration, so during a DST transition (assuming it is happening at 2am), the occurrence will span from 1am to 4am, which still correspond to a duration of 2 hours because of the DST making the time to be 3am when it was 2am. That's how it is coded now. On the contrary, for the second case, we're interested in starting and ending times. With the current coding strategy, the event will span up to 7am the day after, which would be wrong.

Internally, an event can be defined as a duration or with an end time. So we could make a if case here and apply a duration shift if the event is given as a duration or play with the date of the ending time to keep the hour part (could be tricky). But I think we have no way to decide if the event that was entered by the user was intended as a duration or from time to time, from a UI point of view (actually the UI is using the end time paradigm, but we often use it as a duration syntax).

So all in all, I would keep it like that since it does not break the current behaviour and is simple.

pvuorela: > So at least, I'm not introducing any regression here.

That's a good and by such enough for me now :)  -> LGTM

> That being said, could we do better ?

You quite well listed the related problems. The incidences can be assumed either being based on time or duration. The UI could offer both, but then the user would need to be aware of the subtle differences. In theory nice to handle all cases but in practice likely rare case.

PR #4, opened 2021-08-02
Fixup build with new Qt

neochapay:


src/calendarimportmodel.cpp
```diff
@@ -243,7 +227,7 @@ bool CalendarImportModel::importToMemory(const QString &fileName, const QByteArr
             mEventList.append(incidence.staticCast<KCalendarCore::Event>());
     }
     if (!mEventList.isEmpty())
-        qSort(mEventList.begin(), mEventList.end(), incidenceLessThan);
+        std::sort(mEventList.begin(), mEventList.end());
```
dcaliste, line 230: As far as I know, `KCalCore::Incidence` is not defining any `operator<`. I guess that the `QSharePointer` is defining one, based on memory address. So it is compiling, but it will not do the proper comparison for the sort.

dcaliste (changes requested):

pvuorela:
Can't merge and doesn't look like this would be doing proper sorting anyway.

Could accept qSort -> std::sort that keeps the existing comparison function, but that could have a separate PR.

PR #7, opened 2021-09-29
Report when events already exist in db on importation

dcaliste:
Add a "duplicate" role to the import model to be able to flag events that already exist in the database, based on their UID.

Also push two simplification commits:
- the ExtendedStorage::setDefaultNotebook() should not be called. The purpose of this method is to change the flag of a given notebook saying that this notebook is default. It is not the purpose of an importation routine, as far as I can tell.
- filtering the events from all incodences and sorting them can be done easily with the `KCalendarCore::Calendar::events()` routine.

@pvuorela and @chriadam, the root of this MR comes from various discussions in this thread on the forum : https://forum.sailfishos.org/t/bug-4-2-0-21-cannot-import-calendar-entries-via-ics-file/7988/13 The overall goal would be to give a better feedback to the user of what is going on during importation. At the moment, importing twice an event with a UID will result in an importation failed, without explaining to the user what happened. What I'm proposing instead of just a detailed error is to warn the user of possible existing event in the database and let the user overwrite them if they choose to.

This PR requires sailfishos/mkcal#7 since, I'm using the same DEL+ADD pattern than in KCalendarCore.


src/calendarimportmodel.cpp
```diff
@@ -132,50 +140,28 @@ QVariant CalendarImportModel::data(const QModelIndex &index, int role) const
...
-                storage->setDefaultNotebook(notebook);
-            } else {
-                qWarning() << "Invalid notebook UID" << notebookUid;
-                return false;
-            }
+    for (const KCalendarCore::Incidence::Ptr &incidence : mEventList) {
+        const KCalendarCore::Incidence::Ptr old =
+            mStorage->calendar()->incidence(incidence->uid(), incidence->recurrenceId());
+        if (old) {
+            // Unconditionnaly overwrite existing incidence with the same UID/RecID.
```
chriadam, line 158: typo: unconditionally

src/calendarimportmodel.h
```diff
@@ -54,6 +55,7 @@ class CalendarImportModel : public QAbstractListModel
         EndTimeRole,
         AllDayRole,
         LocationRole,
+        DuplicateRole,
```
chriadam, line 58: maybe put this one after the UidRole

src/calendarimportmodel.cpp
```diff
@@ -187,66 +173,39 @@ QHash<int, QByteArray> CalendarImportModel::roleNames() const
...
-            mEventList.append(incidence.staticCast<KCalendarCore::Event>());
-    }
-    if (!mEventList.isEmpty())
-        qSort(mEventList.begin(), mEventList.end(), incidenceLessThan);
 
+    beginResetModel();
+    if (!mEventList.isEmpty())
+        mEventList.clear();
+    mEventList = cal->events(KCalendarCore::EventSortStartDate);
+    for (const KCalendarCore::Event::Ptr &event : mEventList) {
```
chriadam, line 197: Not really important, but I think this will detach due to ranged-for semantics, because mEventList isn't a const ref.

src/calendarimportmodel.cpp
dcaliste: Sorry, thanks for the notice.

src/calendarimportmodel.h
dcaliste: Ok, done.

src/calendarimportmodel.cpp
dcaliste: Yes, we need a `qAsConst` here, but it deosn't exist yet in Qt5.6… I left it like that for simplicity, the size of the list is the number of imported events, so it is small. And the list content are QSharedPointers, so the detach should be cheap : no copy will be done.
I've added a comment for future upgrade though.

chriadam:
LGTM.  I wonder if @blammit or @pvuorela have any comments?

pvuorela:
Didn't test now but seems good to me.

src/calendarimportmodel.cpp
```diff
@@ -187,66 +173,40 @@ QHash<int, QByteArray> CalendarImportModel::roleNames() const
...
-        KCalendarCore::Incidence::Ptr incidence = incidenceList.at(i);
-        if (incidence->type() == KCalendarCore::IncidenceBase::TypeEvent)
-            mEventList.append(incidence.staticCast<KCalendarCore::Event>());
-    }
-    if (!mEventList.isEmpty())
-        qSort(mEventList.begin(), mEventList.end(), incidenceLessThan);
 
+    beginResetModel();
+    if (!mEventList.isEmpty())
+        mEventList.clear();
```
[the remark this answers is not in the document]
dcaliste: You're right. I noticed it also, but wanted to keep the old code there. No problem to remove it. I like removing lines ;-)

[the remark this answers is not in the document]
pvuorela: Which already happened I guess.

src/calendarimportmodel.cpp
```diff
@@ -187,66 +173,38 @@ QHash<int, QByteArray> CalendarImportModel::roleNames() const
...
 
+    beginResetModel();
+    mEventList = cal->events(KCalendarCore::EventSortStartDate);
+    // To avoid detach here, use qAsConst when available.
+    for (const KCalendarCore::Event::Ptr &event : mEventList) {
+        mStorage->load(event->uid(), event->recurrenceId());
+        const KCalendarCore::Event::Ptr old =
+            mStorage->calendar()->event(event->uid(), event->recurrenceId());
+        if (old) {
+            mDuplicates.insert(old->instanceIdentifier());
```
pvuorela, line 201: mDuplicated should be emptied somewhere in this method?

dcaliste: Good catch, I'll update the MR. Thanks.

PR #9, opened 2021-10-11
[nemo-qml-plugin-calendar] Consider time zone for events. JB-43657

dkotin11:


src/calendarworker.cpp
```diff
@@ -754,12 +755,11 @@ CalendarWorker::dailyEventOccurrences(const QList<CalendarData::Range> &ranges,
...
         while (start <= range.second) {
             foreach (const CalendarData::EventOccurrence &eo, occurrences) {
+                QTimeZone tz = QTimeZone::systemTimeZone();
                 // On all day events the end time is inclusive, otherwise not
-                if ((eo.startTime.date() < start
-                     && (eo.endTime.date() > start
-                         || (eo.endTime.date() == start && (allDay.contains(eo.eventUid, eo.recurrenceId)
-                                                            || eo.endTime.time() > QTime(0, 0)))))
-                        || (eo.startTime.date() >= start && eo.startTime.date() <= start)) {
+                const bool isAllDay = allDay.contains(eo.eventUid, eo.recurrenceId) || eo.endTime.time() == QTime(0, 0, 0);
```
pvuorela, line 760: The end time can be 0,0,0 without event being all day?

src/calendarworker.cpp
```diff
@@ -746,20 +748,18 @@ CalendarWorker::eventOccurrences(const QList<CalendarData::Range> &ranges) const
...
+                QTimeZone tz = QTimeZone::systemTimeZone();
                 // On all day events the end time is inclusive, otherwise not
-                if ((eo.startTime.date() < start
-                     && (eo.endTime.date() > start
-                         || (eo.endTime.date() == start && (allDay.contains(eo.eventUid, eo.recurrenceId)
-                                                            || eo.endTime.time() > QTime(0, 0)))))
-                        || (eo.startTime.date() >= start && eo.startTime.date() <= start)) {
+                const bool isAllDay = eo.eventAllDay;
+                if ((isAllDay && eo.startTime.date() <= start && eo.endTime.date() > start) ||
+                   (!isAllDay && eo.startTime.toTimeZone(tz).date() <= start && eo.endTime.toTimeZone(tz).date() >= start)) {
```
dcaliste, line 762: Maybe one can avoid the conversion to the system time zone here with the following:
```
const QDateTime startDt(start); // Will be later start.startOfDay()
...
eo.startTime <= startDt && eo.endTime >= startDt
```

src/calendarworker.cpp
dkotin11: allday events, determined by the flag
Until the fix, the first event in the series has this flag, the rest do not.

pvuorela: Not following. What flag?

src/calendarworker.cpp
dcaliste: Sorry, I was just speaking about the line with the timezone change. Something like:
```
const QDateTime startDt(start); // Will be later start.startOfDay()
const bool isAllDay = eo.eventAllDay;
if ((isAllDay && eo.startTime.date() <= start && eo.endTime.date() > start) ||
    (!isAllDay && eo.startTime <= startDt && eo.endTime >= startDt)) {
```
Like that you're doing QDateTime comparisons and it's guarantee to work if time zones are different. Instead of using the date which requires to adjust the timezones. By creating startDt from start, you're defining 00:00 of that day in the system timezone. Then, you can compare (for non all day events) this startDt with startTime and endTime, even if they are in another timezone.

dkotin11: Seems to understand you.

it won't work that way.
you need two variables.   startDt  and endDt
QDateTime endDt = startDt.addDays(1);

or it is necessary to complicate the condition.  
............(!isAllDay && eo.startTime <= startDt && eo.endTime >= startDt.addDays(1)).....

it doesn't matter to me. It is possible and so and so.   
But the variant with timezone more clearly shows what is happening here and why.

src/calendarworker.cpp
dkotin11: description of errors.
With an event for the all day

use Exchange account

1.  on desctop create allday event
2. I open the device, the event takes two days

1.  on desctop create allday pereodic event
2. open the device, the every event takes two days

if the event is for the whole day, but more than one day.  on device duration is n+1 day

Important. time is not GMT everywhere

pvuorela: You didn't say the timezone of the allday event except that it isn't gmt. For what I see that could be an error in the exchange sync too.

dkotin11: This is a strange exchange problem.
events for the allday are marked as for the allday,
And the time is not important to them, only the date is important.

But when we arrange events by dates, we need to take into account that the event is for the allday (and not look at its time). And for ordinary events, we have to look at the time.

dkotin11: second bug

use the web interface I create an event from 1-00 to 2-00
and from 7-00 to 8-00.

They will be on the computer in one day.
on the device on different days (the first will be on the previous day)

pvuorela: > I am in Moscow. This is UTC + 3

Ok, I interpret this is the "Timezone Z". 

> I have an Exchange account on my work computer and on my phone.
> time zones are configured both there and there 

"Configured..." meaning "Timezone Y" == "Timezone Z"?  Or that the account is configured to the same timezone? 

> (but it is important that it is Exchagne)

No it's not. What is important is what the database contains. 

> create an event for the allday on the computer (use the web interface)

And now still missing most of the information what is the real data. Is that an allday even with clock time (assume no), is it allday with moscow timezone (maybe?).

Please be precise.

dkotin11: for the allday event now.

On desktop timezone UTC+3
on device timezone UTC+3

create allday event on the desktop.


the event arrives on the device with a mark -  "allday" (if it is the first or not periodic).

start event on d 00:00:00   finish on d+1  00:00:00
GMT time,  not UTC+3

trying to expand these events by day.  these days in the local time zone.  
the event gets out for three hours on another day, if you count in time.  But, since it is marked as an all-day event. Then we are only interested in dates.

dcaliste: > start event on d 00:00:00 finish on d+1 00:00:00

Is it information for the serialisation (iCal format ?), or actual values in the DB ? Or the returned values by calling event->dtStart() and event->dtEnd() ?

To get the values in the DB, do :
```select DateStartLocal,StartTimeZone, DateEndDueLocal from components where componentid == 12551;```
And then in a terminal to convert to human readable values :
```date --date=@1633392000```

The serialised data should be:
- DTSTART: 20211014
- DTEND: 20211015

The DB should be:
- DateStartLocal: 20211014T00:00
- DateEndLocal: 20211015T00:00
- StartTimeZone: FloatingDate
(yes, the actual storage for ll day events is exclusive while KCalendarCore is inclusive. This is for backward compatibility reasons and is bad)

The event should be:
- event->dtStart(): 20211014T00:00
- event->dtEnd(): 20211014T00:00
- event->allDay(): true

If it's not the case at any level, the bug is at the level before.

pvuorela:
Conflicting now.

dkotin11:
part of the problem is solved in Daniem request.
Closed it. 

I will recheck, and if something is missing, then we will create a new description of the problem with a new one.

PR #10, opened 2021-10-13
[nemo-qml-plugin-calendar] Save the event when accepting the invitati…

bshmarin:
…on. Contributes to JB#55926

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2014 - 2019 Jolla Ltd.
- * Copyright (c) 2019 - 2020 Open Mobile Platform LLC.
+ * Copyright (c) 2019 - 2021 Open Mobile Platform LLC.
  *
  * You may use this file under the terms of the BSD license as follows:
  *
@@ -166,13 +166,18 @@ bool CalendarWorker::sendResponse(const CalendarData::Event &eventData, const Ca
     const QString &ownerEmail = mNotebooks.contains(notebookUid) ? mNotebooks.value(notebookUid).emailAddress
                                                                  : QString();
 
-    // TODO: should we save this change in DB?
     const KCalendarCore::Attendee origAttendee = event->attendeeByMail(ownerEmail);
     KCalendarCore::Attendee updated = origAttendee;
     updated.setStatus(CalendarUtils::convertResponse(response));
     updateAttendee(event, origAttendee, updated);
 
-    return mKCal::ServiceHandler::instance().sendResponse(event, eventData.description, mCalendar, mStorage);
+    bool sent = mKCal::ServiceHandler::instance().sendResponse(event, eventData.description, mCalendar, mStorage);
+
+    if (!sent)
+        updateAttendee(event, updated, origAttendee);
+
+    save();
+    return sent;
 }
 
 QString CalendarWorker::convertEventToICalendar(const QString &uid, const QString &prodId) const


src/calendarworker.cpp
```diff
@@ -166,13 +166,20 @@ bool CalendarWorker::sendResponse(const CalendarData::Event &eventData, const Ca
...
                                                                  : QString();
 
-    // TODO: should we save this change in DB?
     const KCalendarCore::Attendee origAttendee = event->attendeeByMail(ownerEmail);
     KCalendarCore::Attendee updated = origAttendee;
     updated.setStatus(CalendarUtils::convertResponse(response));
     updateAttendee(event, origAttendee, updated);
+    save();
+
+    bool sended = mKCal::ServiceHandler::instance().sendResponse(event, eventData.description, mCalendar, mStorage);
```
pvuorela, line 175: sent.

src/calendarworker.cpp
```diff
@@ -166,13 +166,20 @@ bool CalendarWorker::sendResponse(const CalendarData::Event &eventData, const Ca
...
+
+    if (!sended) {
+        updateAttendee(event, updated, origAttendee);
+        save();
```
pvuorela, line 179: Regardless of need to save, why do it even twice?

src/calendarworker.cpp
bshmarin: Fixed.

bshmarin:
> > updateAttendee() put the event uid into the mIncidencesToUpdate (SqliteStorage in mkcal). Events which contains in the mIncidencesToUpdate don't fall into the CalendarAgendaModel (by SqliteStorage::Private::addIncidence()). The save() does clean the mIncidencesToUpdate.
> 
> ...and that is a problem because...?

... accepted event is lost from events list in calendar application (until the application restarts).

dcaliste:
Thanks @pvuorela for the ping. I've followed the PR in the last days. I think that saving early is a good idea indeed. Brainstorming a bit more about it, I'm even wondering if we should add to ExtendedCalendar() a "always up-to-date" mechanism. I mean, that any change is triggerring a save of the associated storage (of course we should add something like startUpdates() and finsihUpdates() so we could do save for bunch of changes at once). This up-to-date mechanism should not be the default, but having it in nemo-qml-plugin-calendar would avoid such issues like the one this PR is fixing.

Now about the fact that an ExtendedStorage::load() is overriding any incidences pending for updates or delete looks like a bug to me. At least a design bug. I'm going to think about this to see if something can be done.

dcaliste:
Looking quickly at the code in mKCal, I can read this:
```c++
bool SqliteStorage::Private::addIncidence(const Incidence::Ptr &incidence, const QString &notebookUid)
{
    bool added = true;
    bool hasNotebook = mCalendar->hasValidNotebook(notebookUid);
    // Cannot use .contains(incidence->uid(), incidence) here, like
    // in the rest of the file, since incidence here is a new one
    // returned by the selectComponents() that cannot by design be already
    // in the multihash tables.
    if (isContaining(mIncidencesToInsert, incidence) ||
        isContaining(mIncidencesToUpdate, incidence) ||
        isContaining(mIncidencesToDelete, incidence) ||
        (mStorage->validateNotebooks() && !hasNotebook)) {
```
from sqlitestorage.cpp file in the Private::loadIncidences() function.

So it seems to me that on load, incidences that are not yet synced to the permanent storage are not overridden by the loading process. Maybe there is a real bug (and not a design one), but I need to look closer at the reproducer because just from here, I don't know how it happens.

pvuorela:
No objections for the changes here so let's merge this now at least. Regardless of the bug or what gets loaded by mkcal it's not good if there are pending changes that get written by some unrelated event modification later.

PR #11, opened 2021-11-08
[nemo-qml-pugin-calendar] Fix day definition for events. JB#56431

dkotin11:
The device can be in any time zone.
Events must be arranged by day in order to be displayed correctly.
But the border of days in different zones is different,
so the event must fall on the day of the border of which in the time zone of the device.

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -759,12 +759,9 @@ CalendarWorker::dailyEventOccurrences(const QList<CalendarData::Range> &ranges,
         QDate start = range.first;
         while (start <= range.second) {
             foreach (const CalendarData::EventOccurrence &eo, occurrences) {
-                // On all day events the end time is inclusive, otherwise not
-                if ((eo.startTime.date() < start
-                     && (eo.endTime.date() > start
-                         || (eo.endTime.date() == start && (eo.eventAllDay
-                                                            || eo.endTime.time() > QTime(0, 0)))))
-                    || eo.startTime.date() == start) {
+                const QDateTime startDt(start);
+                if ((eo.eventAllDay && eo.startTime.date() <= start && eo.endTime.date() >= start)
+                    || (!eo.eventAllDay && eo.startTime < startDt.addDays(1) && eo.endTime >= startDt)) {                                        
                     occurrenceHash[start].append(eo.getId());
                 }
             }


chriadam:
Thank you.  Is there some import .ics which we can use to test the issue, or is it possible to add a unit test to exercise this case?

pvuorela:
> The device can be in any time zone. Events are stored in GMT.

The events *may* be stored in GMT/UTC. They can be stored in any timezone or the floating clock time.

The commit message is still much broken. Should have bug reference as JB#43657. Should have empty line after the initial title line. Should have rest of the lines at maximum 72 characters per line.

src/calendarworker.cpp
```diff
@@ -759,12 +759,10 @@ CalendarWorker::dailyEventOccurrences(const QList<CalendarData::Range> &ranges,
...
             foreach (const CalendarData::EventOccurrence &eo, occurrences) {
-                // On all day events the end time is inclusive, otherwise not
-                if ((eo.startTime.date() < start
-                     && (eo.endTime.date() > start
-                         || (eo.endTime.date() == start && (eo.eventAllDay
-                                                            || eo.endTime.time() > QTime(0, 0)))))
-                    || eo.startTime.date() == start) {
+                const QDateTime startDt(start); // Will be later start.startOfDay()
+                const bool isAllDay = eo.eventAllDay;
+                if ((isAllDay && eo.startTime.date() <= start && eo.endTime.date() > start) ||
```
pvuorela, line 764: Operators except comma go to beginning of the line.

pvuorela: Since end time is inclusive for all day events, shouldn't it be ok end date == start date?
The comment could be still worth noting?

dcaliste: Mmh, good point. Since end date is inclusive for all day events, it should be `&& eo.endTime.date() >= start`, to include all day events occurring at `start` date, but also multi-day all-day events including `start` day.

This thing definitely deserve a test, so every cases can be checked (single day, multi-day, strictly before, strictly after, strictly inside, spanning on midnight before, on midnight after, the same for all day cases).

dkotin11: > Operators except comma go to beginning of the line.
done

dkotin11: > Since end time is inclusive for all day events, shouldn't it be ok end date == start date? The comment could be still worth noting?

this is not true for exchange
the moment is not trivial, I left a comment in the code. For exchange, the all-day event ends at 00:00 of the next day

dkotin11: > Mmh, good point. Since end date is inclusive for all day events, it should be `&& eo.endTime.date() >= start`, to include all day events occurring at `start` date, but also multi-day all-day events including `start` day.
> 
> This thing definitely deserve a test, so every cases can be checked (single day, multi-day, strictly before, strictly after, strictly inside, spanning on midnight before, on midnight after, the same for all day cases).

I checked different options.  
first of all, is the Exchange envious or not.
For this, there was a comparison with QTime(0,0)   (

dkotin11:
> > The device can be in any time zone. Events are stored in GMT.
> 
> The events _may_ be stored in GMT/UTC. They can be stored in any timezone or the floating clock time.
> 
> The commit message is still much broken. Should have bug reference as JB#43657. Should have empty line after the initial title line. Should have rest of the lines at maximum 72 characters per line.

done

src/calendarworker.cpp
pvuorela: Exchange is irrelevant here. If kf5-calendarcore defines all-day events to be inclusive of end time, then the end time of the events is included. Period.

If Exchange protocol does something differently, then the sync code needs to adjust the event data it saves.

dcaliste: > this is not true for exchange

I would say that this is a problem of Exchange sync plugin. All fullday events in KCalendarCore are end inclusive. All code using incidences from KCalendarCore must follow this assumption.

If this is not true at this level, I would say that the conversion `Exchange serialisation to KCalendarCore` is buggy for full day events.

> For this, there was a comparison with QTime(0,0)

If I'm not mistaken, the previous comparison with QTime(0,0) was done in the non all day case only. I would say `&& eo.endTime.date() >= start` is enough, but I will be confident only when running tests for it, because it's so easy to make a mistake.

dkotin11: check on QTime(0, 0) was before me.
It allowed it to work with the exchange protocol, without fixes. I left this check. 
fixing the protocol should be part of a different task

dkotin11: correcting the exchange protocol should be a different task.

dcaliste: > check on QTime(0, 0) was before me.
> It allowed it to work with the exchange protocol, without fixes. I left this check.

If I read correctly the condition in its previous form, the test on `QTime(0, 0)` was done after a `eo.eventAllDay ||`, which means that it was taken into account only if the event occurrence does not correspond to an all day event.

In the currently proposed form for the condition, the QTime(0,0) test is appear after a `eo.eventAllDay &&` case, which means it is applied only if the event is a full day one.

So technically, the QTime(0,0) test has different behaviour before and after. If adding it now allows a certain event to work correctly, I'm interested in viewing the iCal serialisation of this event, or let say, know the value of `start`, `eo.startTime` and `eo.endTime` since it may hide a potential unforeseen bug.

The current test, for a given full day event, will return true if:
- its start date is before current,
- and its end date is after current or its end date is current but it has a strictly non 00:00 time.

so it is exactly equivalent to `eo.endTime.date() >= start`, excepting for events where the end timestamp is start with time == 00:00. For instance a full day event with time stamps from 2021-11-09T00:00 to 2021-11-09T00:00 will now rejected, while it is the definition of a full day event for November 9th 2021 (I mean KCalendarCore will return these two time stamps for dtStart() and dtEnd()).

dcaliste: > correcting the exchange protocol should be a different task.

Sure. I agree.

src/calendarworker.cpp
```diff
@@ -759,12 +759,11 @@ CalendarWorker::dailyEventOccurrences(const QList<CalendarData::Range> &ranges,
...
         while (start <= range.second) {
             foreach (const CalendarData::EventOccurrence &eo, occurrences) {
-                // On all day events the end time is inclusive, otherwise not
-                if ((eo.startTime.date() < start
-                     && (eo.endTime.date() > start
-                         || (eo.endTime.date() == start && (eo.eventAllDay
-                                                            || eo.endTime.time() > QTime(0, 0)))))
-                    || eo.startTime.date() == start) {
+                const QDateTime startDt(start);
+                // for exchange, the all-day event ends at 00:00 of the next day
```
pvuorela, line 763: No it does not. Kcalendarcore sets the rules.

dkotin11:
calDAV-services consider this an event for the whole day. The beginning is 0-00, the end is 0-00 of the same day.
These are the ones that I was able to check (and which are often used besides google)

finding for a fix in the sync protocol with exchange.

src/calendarworker.cpp
dkotin11: > > this is not true for exchange
> 
> I would say that this is a problem of Exchange sync plugin. All fullday events in KCalendarCore are end inclusive. All code using incidences from KCalendarCore must follow this assumption.
> 
> If this is not true at this level, I would say that the conversion `Exchange serialisation to KCalendarCore` is buggy for full day events.
> 
> > For this, there was a comparison with QTime(0,0)
> 
> If I'm not mistaken, the previous comparison with QTime(0,0) was done in the non all day case only. I would say `&& eo.endTime.date() >= start` is enough, but I will be confident only when running tests for it, because it's so easy to make a mistake.

done

src/calendarworker.cpp
dkotin11: done

dkotin11:
canceled.
Changes are in https://github.com/sailfishos/nemo-qml-plugin-calendar/pull/13

PR #13, opened 2021-11-18
Fix missing events from different tz in agenda model.

dcaliste:
It's actually the same PR than #11 but I was working on it also based on a new test to ensure that we're not creating regressions, or at least trying to be confident not to !

The test is demonstrating the problem : an event hapenning on D day in Ho Chi Minh city at 3am should be in an AgendaModel featuring D-1 day if this agenda model is created in a Paris time zone.

Creating such test, I actually also found that the AgendaModel is broken for multi-day agendas ! (luckily we're using it only for single day ranges in the calendar application). The `QHash::unite()` has been found to be misleading because uniting a hash with another one that contains the same keys result in the first hash to contains several time the key. Which kind of defeat the initial purpose of a Hash instead of a MultiHash. In Qt5.15, they noticed it and added the `QHash::insert(QHash)` function. Waiting for Qt5.15, I'm proposing to replace the unite with a for loop on single element insertion.

@pvuorela , what do you think ? @dkotin11, sorry for your work on #11, I thought that you started on another bug based on your last comment at that time, when I started this.

### DIFF
--- a/src/calendarmanager.cpp
+++ b/src/calendarmanager.cpp
@@ -809,7 +809,10 @@ void CalendarManager::dataLoadedSlot(const QList<CalendarData::Range> &ranges,
     mLoadedRanges = addRanges(mLoadedRanges, ranges);
     mLoadedQueries.append(instanceList);
     mEvents = mEvents.unite(events);
-    mEventOccurrences = mEventOccurrences.unite(occurrences);
+    // Use mEventOccurrences.insert(occurrences) from Qt5.15,
+    // .unite() is deprecated and broken, it is duplicating keys.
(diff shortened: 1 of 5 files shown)


dkotin11:
> @pvuorela , what do you think ? @dkotin11, sorry for your work on #11, I thought that you started on another bug based on your last comment at that time, when I started this.

These changes completely cover those that were in https://github.com/sailfishos/nemo-qml-plugin-calendar/pull/11.  
We really need them. They solve two problems we found.

chriadam:
LGTM.

pvuorela:
Alright, let's have this.

PR #14, opened 2021-11-18
[nemo-qml-plugin-calendar] Add a displayLabel per occurrence.

dcaliste:
Having a displayLabel per occurrence allows to adjust it for
recurring events like birthdays. So the label can change
with the occurrence and display the age for instance.

@chriadam and @pvuorela , after thinking a lot how to implement an age display for the birthdays, I concluded that it must be done at the occurrence level. It's the object that is changing for a given event (like a day of birth) at every iteration. By overloading the displayLabel (or the description) of the occurrences, one can implement a display of the age at the given day.

Now, there are still some ugly bits :
* we overload the displayLabel depending on the notebook uid, so it is done only for the internal birthday calendar.
* there are two place where the code is duplicated at the moment. I need to clean this.
* I want to avoid as much as possible to introduce translatable strings in the middleware, so the age is given in parenthesis, besides the name, which is nice or not.

Of course, there are one or two places in the Jolla calendar application where there should be a `s/event ? event.displayLabel : ""/occurrence ? occurrence.displayLabel : ""/`. The PR is not done at the moment. I would like to discuss the approach here to the problem first.

### DIFF
--- a/src/calendardata.h
+++ b/src/calendardata.h
@@ -49,6 +49,7 @@ struct EventOccurrence {
     QDateTime recurrenceId;
     QDateTime startTime;
     QDateTime endTime;
+    QString displayLabel;
     bool eventAllDay;
 
     QString getId() const
--- a/src/calendareventoccurrence.cpp
+++ b/src/calendareventoccurrence.cpp
@@ -42,7 +42,20 @@ CalendarEventOccurrence::CalendarEventOccurrence(const QString &eventUid,
                                                  const QDateTime &startTime,
                                                  const QDateTime &endTime,
                                                  QObject *parent)
-    : QObject(parent), mEventUid(eventUid), mRecurrenceId(recurrenceId), mStartTime(startTime), mEndTime(endTime)
+: QObject(parent), mEventUid(eventUid), mRecurrenceId(recurrenceId), mStartTime(startTime), mEndTime(endTime)
+{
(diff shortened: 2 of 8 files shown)


dcaliste:
Some related entries in TJC or in the forum :
* https://forum.sailfishos.org/t/birthday-please-display-persons-age/8994
* https://together.jolla.com/question/166267/add-year-of-birth-or-age-in-calendars-birthday-entries/
* https://together.jolla.com/question/918/display-actual-age-in-birthday-calendar-entries/

I've seen @pvuorela that you commented there, raising the issue of birthday being right on the day and month, but not on the year. That's a valid concern. I've no idea how to treat this properly at the moment.

Another remark made by a user, was about privacy concern : is it polite to have the age displayed beside the name of a person on lock screen (due to event view) ? I don't know. I implemented this PR initially by overriding the event description instead of the display label. Like that the age is only displayed on the detail page.

So some interesting concerns, that deserve some discussions.

pvuorela:
Yea the random year data bothers me. Would be nice if we had a way to set the year undefined on a contact birthday. If only this gets merged I will start getting invalid birthday texts I'm not necessarily even able to make right since I don't know the exact value always, or figuring that out has otherwise little value to me.

Mentioned in the forum post, the ability to open the contact card from the birthday event would be nice, though.

For implementation also a bit pondering would this belong to middleware. Having the specific notebook id is quite a special case. On the UI it would be possible to make better strings, such as that "11th birthday" mentioned in tjc.

PR #15, opened 2021-11-24
[nemo-qml-plugin-calendar] Expose identifier in event list model.

dcaliste:
Modify a bit the event list model, so the identifiers that are the input of the model are also expose as data for each row in the model.

@chriadam , you can see the use for it in https://github.com/dcaliste/harbour/logbook


chriadam:
Also LGTM.

PR #17, opened 2021-12-03
[nemo-qml-plugin-calendar] Update calendar selection algorithm. JB#56605

dkotin11:


pvuorela:
> Update calendar selection algorithm

Strictly speaking this isn't calendar selection algorithm per se here. It's the default notebook property and its definition that's the main thing.

One thing I'm not really sure about is that earlier the default notebook qml property matched the default notebook on mkcal storage side. Now it's mkcal default notebook or unless that's excluded, then the first notebook on the list.

But even if we accepted that the concept of default notebook is not the same, the property is not anymore tracked properly. If the notebook and excluded state changes affect the default notebook, the code should ensure that if the result changes because of them, the defaultNotebookChanged() signal is emitted.

Not sure, but pondering the alternative of updating default notebook status on the app side which wouldn't need change on the definition of the property.

dkotin11:
The properties did not match.
if we did not have the default property set, then the first record of the notebook was selected in qml.
This is the main problem faced by users.  They often turn off their personal notebook.
But if the default notebook was not selected, the disabled personal notebook was returned

PR #19, opened 2022-01-13
[nemo-qml-plugin-calendar] Optimize occurrence loading.

dcaliste:
There are two optimisation done:
- don't reload from storage the recurring events each time a new range is asked for since the recurring mass load is not dependant on ranges. Only reload it on storage change.
- modify the day occurrence bucketing routine. At the moment it is looping over occurrences for every days in ranges. It's better to revert the two loops. Like that on can iterate on days for each occurrences instead on days of ranges.

I have 140 recurring events in my DB at that moment, and this is growing with time. Loading an occurrence is roughly 6 ms on my JollaC. So each time the `loadData()` routine is called in the worker thread, I'm spending 800 ms to (re)load these recurring events, but it has no sense to do it if the storage has not been modified.

I've noticed also that the `dailyEventOccurrences()` routine could be rewritten to save a bit of time. It was doing before D*N operations where D is the number of days in the ranges (typically 30 to 90) and N the number of occurrences in these ranges (typically 40 to 50 in my case). Now it's doing d*N operations where d is the mean duration of one occurrence (typically less than 2 days).

These optimisations help to have a quicker refresh when swiping between months.

Here are the figures for my calendar:
- before
```
[D] unknown:0 - list 149
[D] unknown:0 - date 268
[D] unknown:0 - load 1583
[D] unknown:0 - list 241
[D] unknown:0 - date 469
[D] unknown:0 - load 1920
[D] unknown:0 - list 129
[D] unknown:0 - date 174
[D] unknown:0 - load 1241
[D] unknown:0 - list 131
[D] unknown:0 - date 248
[D] unknown:0 - load 1333
```
- after
```
[D] unknown:0 - list 158
[D] unknown:0 - date 10
[D] unknown:0 - load 1613
[D] unknown:0 - list 240
[D] unknown:0 - date 11
[D] unknown:0 - load 917
[D] unknown:0 - list 120
[D] unknown:0 - date 7
[D] unknown:0 - load 437
[D] unknown:0 - list 120
[D] unknown:0 - date 8
[D] unknown:0 - load 425
```
Each group of 3 figures stands for:
- `list`, the time in ms to load incidences in range from DB,
- `date`, the time to bucket the occurrences per date,
- `load`, the overall time in ms in the worker thread for a `loadData()` call.

The two first groups are for the initial startup of the calendar app, and the two second groups are for two left swipes.

@pvuorela what's your opinion ?


src/calendarworker.cpp
```diff
@@ -784,7 +780,7 @@ void CalendarWorker::loadData(const QList<CalendarData::Range> &ranges,
     foreach (const QString &id, instanceList)
         mStorage->loadIncidenceInstance(id);
 
-    if (!ranges.isEmpty()) {
+    if (!ranges.isEmpty() && reset) {
```
pvuorela, line 783: Trying this with my calendar that has a bunch of repeating events per week, it doesn't seem to load them. Didn't go investigating more than removing this addition gets them back, but at least the mResetPending on manager is initialized false.

pvuorela:
Noticed some problem but in general the rationale sounds sane. Though could include that in the commit message too.

src/calendarworker.cpp
dcaliste: You're right, I didn't pay enough attention. Actually, I've some excluded notebooks, so the `mResetPending` was set to `true` on start up in my case when the excluded notebook setting is read.

I think it's safe to initialise `mResetPending` to `true` to get the first initial load to run without other side effects. I've amended the commit in that sense.

dcaliste:
> Though could include that in the commit message too.

I've developped the commit message to be more clear in case the modifications should be looked at later on in the future.

src/calendarworker.cpp
pvuorela: Hm, should this isEmpty() check be here actually? Thinking what happens if e.g. notebooks get loaded on startup, excluded notebooks trigger reset/loadData from the manager and this triggers without ranges set for some reason.

src/calendarworker.cpp
```diff
@@ -752,23 +752,19 @@ CalendarWorker::eventOccurrences(const QList<CalendarData::Range> &ranges) const
...
-                // On all day events the end time is inclusive, otherwise not
-                if ((eo.eventAllDay && eo.startTime.date() <= start
-                     && eo.endTime.date() >= start)
-                    || (!eo.eventAllDay && eo.startTime < startDt.addDays(1)
-                        && eo.endTime >= startDt)) {
-                    occurrenceHash[start].append(eo.getId());
-                }
+    for (const CalendarData::EventOccurrence &eo : occurrences) {
+        const QDate st = eo.eventAllDay ? eo.startTime.date() : eo.startTime.toLocalTime().date();
+        const QDate ed = eo.eventAllDay ? eo.endTime.date() : eo.endTime.toLocalTime().date();
```
pvuorela, line 760: Not sure I'm following. Earlier there was the allday inclusivity check but how is toLocalTime() related ot that or why such is needed?

dcaliste: When I exchange the `for` loops the previous inner test was :
```
   if ((eo.eventAllDay && eo.startTime.date() <= start
                     && eo.endTime.date() >= start)
                    || (!eo.eventAllDay && eo.startTime < startDt.addDays(1)
                        && eo.endTime >= startDt)) {
                    occurrenceHash[start].append(eo.getId());
                }
```
with `start` being the date when the occurrence may happen or not.

If I read this `if` condition properly, it says:
* in case of all day events, `start` is taken into account if the occurrence first day is at `start` or before and also if the occurrence last day is at `start` or later.
* in case of non all day events, `start` is taken into account if the start time in the local time zone is during `start` date or earlier, and if the end time in the local time zone is later than the start of day of `start`.

I reverse the `for` loop order so the inner one is now iterating on dates the occurrence is spanning over, following the same convention as the `if` before, namely :
* for all day events, the start date is taken as the `date()` of the startTime, without paying attention to the time zone, like before, and spanning up to the `date()` of endTime. I think doing like that, I'm respecting the `<=` and `>=` that were previously there.
* for non all day event, the date must be taken in the local time zone, thus the conversion.

The agendamodel test supposed to check this is actually working with the patch. I was not sure myself of not having done any mistake when simplifying the routine like that, but it seems ok. That being said, an addition look and explanation is of course not a wasted time. Tell me if you agree with my reasoning.

dcaliste: Thinking more about this, in particular about the mention that non all day are end exclusive, I now believe that the previous `if` was a bit erroneous. Indeed for the all day events, the test on end time is done with a `>=`, but it is also done like that for non all day event. I think it should have finished with `&& eo.endTime > startDt))`.

I've just amended the commit to add a `.addSecs(-1)` to the generation of the date in the non all day case. I've readded the comment also to make it clear.

With it, I can create an event from 10pm to 00:00 the day after, and having it displayed only on the correct day and not having a 00:00 to 00:00 occurrence also the day after in the calendar.

Thanks a lot for noticing this and warning me.

dcaliste:
I've amended the commit to use a flag in the worker to know if the recurring events have already been loaded or not. Doing it also allow to revert the change in the manager on the initialisation of `mResetPending`, which is a nice side effect !

src/calendarworker.cpp
pvuorela: Yea, the the inclusive / exclusive looks better now. Resulting times from the event being always inclusive.

The remaining detail I'm pondering with this toLocalTime difference is all-day events with timezone set to non-current timezone. Those are tricky, likely not what the user expects and that's why the Calendar app creates all-day events without timezone. But remotely synced event could contains such. E.g. I'm in Finland and there's a 1 one day all-day event with some US timezone. That should end up shown on two days, right?

dcaliste: The problem of time zone for full day events is a tricky one and I think I'm keeping the previous behaviour here. Namely, we convert the occurrence date time (start or end) into a date without paying attention to the time zone it was defined into (see previous code like `eo.startTime.date() <= start`, where `start` is a date itself.

We discussed this a long time ago many time on IRC with @chriadam and we arrive to the conlusion that :
* case DTSTART is a DATE, like in DTSTART:20220120 cannot has a time zone definition (see https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.3.1). It is always resolving as a full day whatever the time zone. For instance, my example will start at 20220119T180000Z when you are in Ho Chi Minh city (UTC+06).
* case DTSTART is a DATE-TIME, without time zone and starting at 00:00 local time, like DTSTART:20220120T000000 (duration is supposed to be a multiple of a day). mKCal is flagging such as full day events (see sqliteformat.cpp#1216).
* case DTSTART is a DATE-TIME with a time zone and starting at 00:00 in this time zone, like DTSTART;TZID=Asia/Ho_Chi_Minh:20220120T000000. In that case, we are considering this event as _not_ being full day, even in Viet Nam time zone. It is just an event that happens from midnight. It is not flagged as full day by mKCal.

The two first cases are internally stored by KCalendarCore (and restored by mKCal) as an event `e` with `e.fullDay() == true` and `e.dtStart().time() == 00:00` and `e.dtStart().spec() == LocalTime`. Thus, with this patch (and also without) they are taking the fullDay branch in the condition and the date is taken without paying attention to the time zone (there is not any anyway, spec being LocalTime).

The last case is stored internally by KCalendarCore (and restored by mKCal) as an event `e` with `e.fullDay() == false` and `e.dtStart().time() == 00:00` and `e.dtStart().spec() == TimeZone`. Then, it will enter the branch for non full day and the date will be extracted _after_ a conversion to the local zone of the device (and may thus be visible during two days on the device if the local time zone is different from the one of the event).

I think that this is thus respecting the three cases cited above.

pvuorela: Fair enough. I could consider it a bug in mkcal if it restores an event different than what was saved to the database, e.g. losing the allday flag if there was one earlier. But that could be a different thing and maybe not the highest priority thing to revisit now.

pvuorela (approved):
Indeed the recurrence loading looks safer now and also easier to follow. All good, thanks!

PR #20, opened 2022-01-18
[nemo-qml-plugin-calendar] Expose the event status to QML.

dcaliste:
It is exposing the STATUS property to QML, but it requires a patch in KCalendarCore that is not yet there, since KCalendarCore is currently hiding cancelled occurrences of recurring events.

It can be tested with plain cancelled events though.


pvuorela:
Already showing me nicely the cancelled event info on my exchange calendar. LGTM

pvuorela (approved):
Let's have this already.

PR #21, opened 2022-01-24
[nemo-qml-plugin-calendar] Add cancelled property in lightweight. Con…

dcaliste:
…tributes to JB#56640

It's a continuation of #20. To simplify DBus transfert between the daemon and the client, I decided to just expose the cancel status and not all possible statii. We can still rework this if we would like to support other values for status.

@pvuorela may you review this when you have the time ?


pvuorela (approved):
Nice!

PR #22, opened 2022-01-28
[nemo-qml-plugin-calendar] Compare also time spec when assigning star…

dcaliste:
…t or end times.

The QDateTime::operator== used to check if start or end time of
an event modification has been changed is comparing points in time
while one may be interested to change the time spec (from a time
zone one to a floating one), resolving locally to the same point
in time.

Reproducer:
- create an event with a given start date time in the system time zone (default),
- modify the event, keep the time, but change to local time (no time zone).
(@pvuorela, actually, if you look at the event modification page again, it is still stored with the system time zone, but that's an issue with mKCal, to see that the proper spec is passed to mKCal, add a debug message in the worker thread displaying `dtStart` for the saved event in `::saveEvent()`)

It's funny that KCalendarCore guys were bitten by the same bug one month ago, see https://invent.kde.org/frameworks/kcalendarcore/-/commit/9f4628aa2c6c90c94fd36d2d4ac8e940adc93680 So to see the fix working here, you also need to use the latest 5.90 version of KCalendarCore.

The fix is not taking into account cases where one would actually change the time zone and the time itself, so the old and the new point to the same point in time… Do you think it's worth to add it, or it's overly complicating things ?

### DIFF
--- a/src/calendareventmodification.cpp
+++ b/src/calendareventmodification.cpp
@@ -108,7 +108,10 @@ void CalendarEventModification::setStartTime(const QDateTime &startTime, Qt::Tim
 {
     QDateTime newStartTimeInTz = startTime;
     updateTime(&newStartTimeInTz, spec, timezone);
-    if (m_event.startTime != newStartTimeInTz) {
+    if (m_event.startTime != newStartTimeInTz
+        || m_event.startTime.timeSpec() != newStartTimeInTz.timeSpec()
+        || (m_event.startTime.timeSpec() == Qt::TimeZone
+            && m_event.startTime.timeZone() != newStartTimeInTz.timeZone())) {
         m_event.startTime = newStartTimeInTz;
         emit startTimeChanged();
     }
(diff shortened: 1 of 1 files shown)


dcaliste:
> to avoid a change signal.

Well, actually it's even a bit more than a missed signal. If you scheduled a meeting with London friends at 5pm French time, but decided that it's better if saved as 4pm UK time, you cannot do it. Because the m_event.startTime won't be updated because the spec didn't change and the point in time neither. So your event will always be seen as 5pm Paris time. Not a big deal in my opinion. Just mentioning.

pvuorela:
Hm. Was assuming QDateTime != being able to detect timezone changes as it promises "Two datetimes are different if either the date, the time, or the time zone components are different" but the actual implementation doesn't seem to meet that.

Maybe we should have some extra check to compare the timezones too to make it robust? The change signal indeed doesn't matter much but the skipped assignment might bite back.

dcaliste:
I've added the test for change in the time zone also, in case. But actually, this will not work out of the box and requires to patch KCalendarCore also since there is a call to `->setDtStart(eventData.startTime)` in the worker thread that will not update the internal value of dtStart.

But at least from our side we're safe. Too bad that Qt implements the operator!= like !operator==. An update to the documentation would be nice (I guess a fix in the code would break too many things).

pvuorela (approved):
Thanks. Looking good and at least more robust on our side.

PR #23, opened 2022-02-09
[nemo-qml-plugin-calendar] Save EXDATE in localtime only if the event…

dcaliste:
… is in localtime.

@pvuorela, this is a follow-up of sailfishos/mkcal#13 that introduced a "strange" behaviour. Indeed, QDateTime coming from QML are always in `Qt::LocalTime`, so now they are stored as that and not converted to the system time zone. So when deleting an occurrence of a recurring event, an EXDATE is added as `EXDATE:20220215T0900` while the event is given with a time zone. So the exdate is taken into effect only if the device is in the right time zone. Changing time zone makes the deleted occurrence appearing again. And further more, while being RFC valid, this is confusing some caldav servers (openexchange for instance) if there are exdates in local time and exdates with a time zone.

I'm proposing to enforce EXDATE in the same spec or time zone than the DTSTART.
Since it's RFC valid to mix local time and time zone, I didn't enforce this at mKCal level, but prefer to do it here.

### DIFF
--- a/src/calendarutils.cpp
+++ b/src/calendarutils.cpp
@@ -259,38 +259,46 @@ QList<QObject *> CalendarUtils::convertAttendeeList(const QList<CalendarData::At
 }
 
 CalendarData::EventOccurrence CalendarUtils::getNextOccurrence(const KCalendarCore::Event::Ptr &event,
-                                                               const QDateTime &start)
+                                                               const QDateTime &start,
+                                                               const KCalendarCore::Incidence::List &exceptions)
 {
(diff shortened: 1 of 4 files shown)


dcaliste:
I've just added 3 commits to update tests since some were broken in master branch due to earlier modifications in mkcal or kcalendarcore.

dcaliste:
I've just amended a bit the initial commit to add a test for it demonstrating the problem.

pvuorela (approved):
Seems good to me.

PR #25, opened 2022-02-21
[nemo-qml-plugin-calendar] Factorise code between classes sharing eve…

dcaliste:
…nt characteristics.

@pvuorela , this is a proposition to factorise a bit of code in nemo-qml-plugin-calendar.

Create a base class Event, based on CalendarData::Event struct,
from which inherit:
- a StoredEvent class for events owned by a Manager,
- a EventModification class to create new events or
  modify data from StoredEvent instances,
- a ImportEvent class for events read from ICS data
  and not stored on device.

These classes mainly reuse the existing code or
class layouts.

All code handling transformations from an to
KCalendarCore::Event have been gathered and moved
to the CalendarData::Event structure.

This last part is to avoid to have this transformation code scaterred in utils and worker, like it was before. Often, when adding a new member in CalendarData::Event structure to mirror one from KCalendarCore, I forgot to put it in utils, so it can be seen from CalendarImportEvent also for instance. And when moving away from CalendarData::Event in favour of KCalendarCore::Incidence, the affected code will be less scattered also.

What do you think ? As a bonus, it is reducing the code volume by 200 lines ;-)

### DIFF
--- a/src/calendardata.h
+++ b/src/calendardata.h
@@ -38,7 +38,7 @@
 #include <QDateTime>
 
 // KCalendarCore
-#include <KCalendarCore/Attendee>
+#include <KCalendarCore/Event>
 
 #include "calendarevent.h"
 
@@ -66,20 +66,25 @@ struct Event {
     bool readOnly = false;
     bool rsvp = false;
     bool externalInvitation = false;
-    CalendarEvent::Recur recur;
+    CalendarEvent::Recur recur = CalendarEvent::RecurOnce;
     QDate recurEndDate;
     CalendarEvent::Days recurWeeklyDays;
-    int reminder; // seconds; 15 minutes before event = +900, at time of event = 0, no reminder = negative value.
+    int reminder = -1; // seconds; 15 minutes before event = +900, at time of event = 0, no reminder = negative value.
     QDateTime reminderDateTime; // Valid when reminder is at a given date and time.
     QString uniqueId;
     QDateTime recurrenceId;
     QString location;
-    CalendarEvent::Secrecy secrecy;
+    CalendarEvent::Secrecy secrecy = CalendarEvent::SecrecyPublic;
     QString calendarUid;
     CalendarEvent::Response ownerStatus = CalendarEvent::ResponseUnspecified;
     CalendarEvent::Status status = CalendarEvent::StatusNone;
     CalendarEvent::SyncFailure syncFailure = CalendarEvent::NoSyncFailure;
     CalendarEvent::SyncFailureResolution syncFailureResolution = CalendarEvent::RetrySync;
+
+    Event() {}
+    Event(const KCalendarCore::Event &event);
+
+    void toKCalendarCore(KCalendarCore::Event::Ptr &event) const;
 
     bool operator==(const Event& other) const
     {
@@ -90,6 +95,14 @@ struct Event {
     {
         return !uniqueId.isEmpty();
     }
+
+private:
+    int fromKReminder(const KCalendarCore::Event &event) const;
+    QDateTime fromKReminderDateTime(const KCalendarCore::Event &event) const;
+    void toKReminder(KCalendarCore::Event &event) const;
+    CalendarEvent::Days fromKDayPositions(const KCalendarCore::Event &event) const;
+    CalendarEvent::Recur fromKRecurrence(const KCalendarCore::Event &event) const;
+    void toKRecurrence(KCalendarCore::Event &event) const;
 };
 
 struct Notebook {
--- a/src/calendarevent.cpp
+++ b/src/calendarevent.cpp
@@ -38,28 +38,38 @@
 
 #include "calendarutils.h"
 #include "calendarmanager.h"
+#include "calendardata.h"
 
-CalendarEvent::CalendarEvent(CalendarManager *manager, const QString &uid, const QDateTime &recurrenceId)
-    : QObject(manager), mManager(manager), mUniqueId(uid), mRecurrenceId(recurrenceId)
+CalendarEvent::CalendarEvent(const CalendarData::Event *data, QObject *parent)
+    : QObject(parent), mData(new CalendarData::Event)
 {
-    connect(mManager, SIGNAL(notebookColorChanged(QString)),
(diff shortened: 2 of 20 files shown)


src/calendardata.h
```diff
@@ -90,6 +205,200 @@ struct Event {
...
+            }
+            case CalendarEvent::RecurYearly:
+                event.recurrence()->setYearly(1);
+                break;
+            case CalendarEvent::RecurCustom:
+                // Unable to handle the recurrence rules, keep the existing ones.
+                break;
+            }
+        }
+    }
```
pvuorela, line 401: Earlier this was mostly data type definition file now there's a bunch of code too. Could create a .cpp file for the methods?

src/calendarevent.cpp
```diff
@@ -89,168 +99,228 @@ static Qt::TimeSpec toTimeSpec(const QDateTime &dt)
...
 QString CalendarEvent::uniqueId() const
 {
-    return mUniqueId;
+    return mData->uniqueId;
 }
 
 QString CalendarEvent::color() const
 {
-    return mManager->getNotebookColor(mManager->getEvent(mUniqueId, mRecurrenceId).calendarUid);
+    return QString();
```
pvuorela, line 162: Should this be in the base class if it doesn't return anything? Maybe at least could warrant some comment here.

src/calendareventmodification.h
```diff
@@ -41,34 +41,27 @@
 #include "calendarchangeinformation.h"
 #include "calendarcontactmodel.h"
 
-class CalendarEventModification : public QObject
+class CalendarEventModification : public CalendarEvent
```
pvuorela, line 44: Was assuming for a moment that these are redundant Q_PROPERTYs but then realized there's also WRITE methods declared.  Perhaps that could also warrant some note.

src/calendarimportevent.h
```diff
@@ -38,9 +38,10 @@
...
 #include <KCalendarCore/Event>
 
+#include "calendardata.h"
 #include "calendarevent.h"
 
-class CalendarImportEvent : public QObject
+class CalendarImportEvent : public CalendarEvent
 {
     Q_OBJECT
     Q_PROPERTY(QString displayLabel READ displayLabel CONSTANT)
```
pvuorela, line 47: Ponder if this set of redeclarations should be included. Feels interesting when subclass makes the properties constant but guess it's not necessarily wrong per se. Practical effect shouldn't be big when importing is rare and limited set of events.

src/calendareventmodification.h
dcaliste: I've added a note mentioning the overloading with the write definition.

src/calendarimportevent.h
dcaliste: Yeh, I didn't want to change the previous implementation, but I agree that there is no point in declaring things constant here. I mean they are, but the performence delta in real life is neglectable I think. So I've removed the overloading declaration here.

src/calendardata.h
pvuorela: Hm, could it be a middle ground if the code would reside now in calendarutils.cpp instead of a separate file? That used to have parts of it already earlier.

dcaliste: Sure, I've moved all the `Event`code into `calendarutils.cpp`.

dcaliste:
I've rebased and corrected the compilation issue in the tests.

src/calendarevent.cpp
```diff
@@ -89,168 +99,223 @@ static Qt::TimeSpec toTimeSpec(const QDateTime &dt)
...
+
+    return mManager->convertEventToICalendarSync(mData->uniqueId, prodId);
+}
+
+QString CalendarStoredEvent::color() const
+{
+    return mManager->getNotebookColor(mData->calendarUid);
+}
+
+void CalendarStoredEvent::setEvent(const CalendarData::Event *data)
```
pvuorela, line 283: Hm, could this be a const reference here? Would avoid any interpretation that this would involve any instance ownership etc.

src/calendarevent.h
```diff
@@ -205,10 +197,35 @@ private slots:
     void rsvpChanged();
     void externalInvitationChanged();
 
+protected:
+    CalendarData::Event *mData;
```
pvuorela, line 201: Any reason to have this dynamically allocated or could it be without the pointer part?

src/calendareventmodification.cpp
```diff
@@ -55,129 +55,97 @@ void updateTime(QDateTime *dt, Qt::TimeSpec spec, const QString &timeZone)
 
 }
 
-CalendarEventModification::CalendarEventModification(CalendarData::Event data, QObject *parent)
-    : QObject(parent), m_event(data), m_attendeesSet(false)
+CalendarEventModification::CalendarEventModification(const CalendarStoredEvent *source, QObject *parent)
+    : CalendarEvent(source)
 {
+    if (parent)
+        setParent(parent);
```
pvuorela, line 62: Why separate setParent() and not as parameter to base class?

src/calendarevent.cpp
dcaliste: The problem here is a cross definition : CalendarData::Event is using enums defined in CalendarEvent class.

To avoid this difficulty, I'm using a pointer. It's ugly. But transitory only. Then, when moving to KCalendarCore::Incidence::Ptr (since dropping CalendarData::Event completely), we can use references.

src/calendarevent.h
dcaliste: It's the same reason here. I don't like it neither, but I had no idea how to easily circumvent, without creating an addition file for the enum definitions.

src/calendareventmodification.cpp
dcaliste: Good question. Trying to add the `QObject *` argument to the `CalendarEvent` copy constructor, I noticed that I did so because I misinterpreted the compiler error at that time telling that `nullptr` was ambiguous. I thought it was unable to distinguish `Calendar::Event*` to `CalendarEvent*` seeing to `void*`. But no of course. So I changed the `nullptr` by casting `0` to the right type.

It's using proper `QObject*` argument now.

pvuorela (approved):

PR #28, opened 2022-03-04
[nemo-qml-plugin-calendar] Don't reload from DB when not needed.

dcaliste:
The follow-up PR from sailfishos/mkcal#17 implementing the `storageUpdated` signal.

Basically, the major change is that when this signal is received, we're not flushing loaded ranges and notebooks, so on next `loadData()` call from the manager, we're not going to poke the DB. The rest is the same. I'm not taking any advantage of the added, modified or deleted lists at the moment. I don't know if it could be worth to do it in the future or not. For sure, it would add more complexity because we would need to get the internal storage of CalendarManager in sync. Maybe later when it will use KCalendarCore storage directly.

@pvuorela, it's still a bit WIP and I haven't tested it for real on device yet. But you may already give some feedback on design if you have time to think about it.

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -72,7 +72,7 @@ namespace {
 }
 
 CalendarWorker::CalendarWorker()
-    : QObject(0), mAccountManager(0), mHasRecurringEvents(false)
+    : QObject(0), mAccountManager(0)
 {
 }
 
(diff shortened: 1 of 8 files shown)


src/calendarworker.cpp
```diff
@@ -87,32 +87,31 @@ CalendarWorker::~CalendarWorker()
...
+    Q_UNUSED(modified);
+    Q_UNUSED(deleted);
 
-void CalendarWorker::storageFinished(mKCal::ExtendedStorage *storage, bool error, const QString &info)
-{
-    Q_UNUSED(storage)
-    Q_UNUSED(error)
-    Q_UNUSED(info)
+    // No smart handling of known updates at the moment.
+    // mSentEvents is reseted to reflect the fact that the manager reseted mEvents itself.
```
pvuorela, line 112: Nit: reseted -> reset

dcaliste: Corrected, thanks.

dcaliste:
As far as I'm concerned, it's good to go. Hopefully it will not break too many things… Quick usage on my phone didn't trigger something bad, but got immediate speed improvements when modifying, deleting or adding an event in the calendar application.

dcaliste:
Regarding the discussion in sailfishos/mkcal#19, the solution proposed there could be applied before or after this PR. I think, mostly after to let the time to review properly  sailfishos/mkcal#19.

pvuorela:
Compiled the mkcal PR and this and tried on the device first unit tests. Mkcal goes fine but here I start to get couple failures.

```
FAIL!  : tst_CalendarManager::test_notebookApi() Compared values are not the same
   Actual   (mManager-&gt;defaultNotebook())   : &quot;20580b38-c428-4f82-a97a-e7e93a274628&quot;
   Expected (mAddedNotebooks.first()-&gt;uid()): &quot;6f9d3952-ef87-473c-83b2-ed266aa9a56f&quot;
   Loc: [tst_calendarmanager.cpp(596)]

FAIL!  : tst_CalendarEvent::testAttendees() Compared values are not the same
   Actual   (attendees.count()): 4
   Expected (3)                : 3
   Loc: [tst_calendarevent.cpp(733)]
```

dcaliste:
Sorry for the issues. I should have messed-up with `LD_LIBRARY_PATH` when testing.

I'm looking ito it but not everything is solved at the moment. So the additional commit is still WIP.

pvuorela (approved):

PR #30, opened 2022-03-10
Use KCalendarCore::Incidence::Ptr as a storage container

dcaliste:
Completely remove `CalendarData::Event` structure to use `KCalendarCore::Incidence::Ptr` instead.

The advantages of this are:
- all unsupported features up to now are now available for QML exposure, like todos, journals, multi alarms or attachments,
- less code duplication with KCalendarCore (we gain an additional reduction of 200~250 lines),
- no more race conditions between worker and manager between main event data and attendee data,
- less memory consumption since the worker and the manager share the same data.

This last point is crucial and holding on some code assumptions:
- the `KCalendarCore::Incidence::Ptr` provided by the manager must be _read-only_. This is somehow guaranteed by the new API separation between `CalendarEvent` and `CalendarStoredEvent`. The `CalendarEventModification` are always built on a clone of the data. The reason for this is, if any modification is done to the data pointed by `KCalendarCore::Incidence::Ptr` from a `CalendarStoredEvent::mIncidence` these modifications will be spuriously saved to the database at the next `save()` call.
- mKCal should not modify any data pointed by `KCalendarCore::Incidence::Ptr` objects owned by the `ExtendedCalendar` object of the worker. This is how mKCal is built at the moment, so it should not be a problem. If the storage is modified, the calendar is emptied before being repopulated (and not modified in place). So the shared pointers exposed in the UI will not be shared anymore and only owned by the UI, up to the moment, the manager is reseted also after receiving `dataLoaded`.


dcaliste:
Post-scriptum, the PR is made of many small commits. It's how the modifications have been done, touch by touch. At each commit, the tests are passing. I think it's also easier for review to look a the steps themselves. But when it will be done, commits could be squashed, or not. As you prefer, @pvuorela.

dcaliste:
Rebase on top of master.

Also add a test to check that CalendarStoredEvent object is properly emitting signal on content change. This was failing in the previous form after rebase. It was due to the fact that I didn't respect the rule I wrote in introduction :/ When saving modifications in the worker thread, the KCalendarCore::Incidence::Ptr target owned by the manager is touched by the worker before saving... I solved the problem in a general way, by cloning the incidence for these cases :
- in an event query,
- in a CalendarStoredEvent.
I think it's not much of additional memory consumption, because there are only a few number of these objects used in the UI. So the cloning make their content totally safe without too much memory pressure. The manager for instance is still sharing memory with the worker for all events.

pvuorela:
Sorry, but from thread-safety point of view I'm not sure about these. The incidences themselves don't have any mutex type of locking and the Ptr types are QSharedPointer which states: 

"QSharedPointer and QWeakPointer are reentrant classes. This means that, in general, a given QSharedPointer or QWeakPointer object cannot be accessed by multiple threads at the same time without synchronization."

Having the common thread side read-only wouldn't seem enough if at the same time the worker is allowed to do changes. Some mutex could provide safety but then the benefits of threading are easily gone. Or (even if the kf5-calendarcore documentation doesn't promise anything on thread-safety) limiting the access to const methods without locking probably works it could lead to more complicated code.

So think it could be better if the threads don't share any instances of these types. That could be avoided by clone() on everything that's passed to the other thread if the sender needs to keep a reference, but then there would go the memory consumption wins.

There are some commits in the beginning that seem good: the unit test, the Person usage seems safe, and the sendResponse. And no objections in switching Events to Incidences either.

dcaliste:
I thank you for the points you raised on threading-related possible issues. Let review them:

> a given QSharedPointer or QWeakPointer object cannot be accessed by multiple threads

I agree. Let me rephrase to see if I understood it well. We cannot read an incidence pointed by a QSharedPointer in the manager, while the worker thread is modifying this incidence. All other cases are safe, including, both threads reading part of the incidence at the same time, or even the worker clearing its `::Ptr` while the manager is still using it (read or store, the manager will become the owner of the data then).

We have two possibilities:
- clone everything when passing from the worker to the manager in the `loadData()` method. We then loose the memory gain.
- try to see every places where the worker could modify the incidences pointed by any `::Ptr` shared with the manager and ensure that we handle these cases safely (reverse need also to be checked : that the manager is not doing any change on an incidence shared with the worker, but as far as I know, the manager is not doing any modification itself on data).

I'm still trying to achieve the second possibility, with the option to fallback to the first if I'm failing.

Reviewing the code, the only places I can spot in the worker where an incidence is modified are:
- in `saveEvent()` within the `startUpdate()` and `endUpdate()` calls.
- in `deleteEvent()` when the recurrence is modified to add an exdate.
- (actually, there are two other places, `dissociateSingleOccurrence()` and `sendResponse()`, but they are called in a blocking state, so there is no concurrency involved).

I'm not in favour of locking, as you mentioned, otherwise we would loose thread concurrency. What I'm proposing is to clone on the manager side. I've added a `detachIncidence()` routine to do it. The idea is based on the fact that the worker is not doing any modification without the manager asking for them. So we know before poking the worker what will be sensitive data. So we clone these incidences in order to the manager to have a separated version of what the worker thread will work on. Like that during the time the worker is touching the incidences, the manager can receive requests from the UI objects (like an event query) to get CalendarStoredEvent objects or to refresh an event query ; and honours these requests with safe data that won't be touched by the worker thread. When the worker has finished its work, the storageModified signal will be emitted normally and the manager will ask for new data and drop the cloned versions when getting the new data from the worker.

These changes, with the previous ones ensuring that all UI objects (CalendarStoredEvents or CalendarEventQuery) are working on clones, should prevent us from threading issues while keeping the memory demand to the minimum.

> There are some commits in the beginning that seem good: the unit test, the Person usage seems safe, and the sendResponse. And no objections in switching Events to Incidences either.

Do you prefer to get these commits in a separated PR, so changes in this PR will only be related to the switch to KCalendarCore::Incidence as a storage ?

What I can do also is to implement the switch using option 1 (clone in the worker thread before pushing data to the manager) and add on top of that a commit that try to aleviate the memory consumption with the sharing option (plus detach) ? Notice that the memory consumption is currently double due to the fact that the worker is storing all incidences in the `mCalendar` variable, while the manager has a copy of all of them in `mEvents` variable, in term of `CalendarData::Event` (which is less than a KCalendarCore::Incidence of course).

pvuorela:
> I agree. Let me rephrase to see if I understood it well. We cannot read an incidence pointed by a QSharedPointer in the manager, while the worker thread is modifying this incidence. All other cases are safe, including, both threads reading part of the incidence at the same time, or even the worker clearing its ::Ptr while the manager is still using it (read or store, the manager will become the owner of the data then).

I read the Qt documentation as discouraging use of anything related to QSharedPointer on multiple threads without a locking mechanism. Reading the pointer value is probably ok in practice, but I wouldn't go trusting that e.g. copying such a Ptr instance wouldn't cause problems as that should make the reference counting to kick in, and that seems like not implemented thread-safely on this type.

That would mean that we shouldn't use Ptr as method parameter type or on the return value types (i.e. only pointers or references to Ptr). Also the Incidence::List and similar non-pointer/reference containers should be banned for anything that's shared between the threads. And probably should avoid any methods that involve these in mkcal and kf5-calendarcore. Would feel nearly impossible to keep done correctly and robustly so think it's better if we just avoid sharing any instances between the threads.

That leaves us the option of clone()ing everything that's passed to the other thread if the sender keeps a reference. Whether that makes sense I'm not sure. The classes we have currently use are lighter as we only need to pass the information that's needed. Qt base type copy-on-write probably saves something on the clone()s but then again some of that probably gets implicitly used on our classes too. My feeling is that there isn't much to win and it would come at the expense of robustness or understandability, unfortunately.

> Do you prefer to get these commits in a separated PR, so changes in this PR will only be related to the switch to KCalendarCore::Incidence as a storage ?

Separate PR would be a good way to get some parts in.

pvuorela:
> Separate PR would be a good way to get some parts in.

To add, imo the unit test and the response api adjustment would be the best mergeable. 

The Person and the Event -> Incidence border line. Former redundant type and seems safe, though it's not either too many lines and as general rule the intention has been on minimizing kcalcore on UI thread for avoiding unexected threading issues. Latter per se good but guess it's helpful mostly if we add support for journals and todos etc for which i'm skeptical if would be adding support.

dcaliste:
Thank you @pvuorela for the discussion here. I think you convinced me that sharing ::Ptr is not a good idea. I based my reasoning on the faulty assumption that ref counting was done thread safe. Without safe ref counting, this is hopeless.

I'm going to separate the commits that looks unrelated, as you mentioned.

I'm still convinced though that moving to KCalendarCore structures instead of duplicating everything is the way to go forward. Thinking more about the thread issue, I ended up with the conclusion that if we don't want any memory duplication between thread and manager, the thread should not store anything. I've begun to think of an alternate implementation :
- keep all the KCalendarCore::Incidence::Ptr in the manager (maybe even make the manager being a MemoryCalendar, so sort by date and so on may be reused from upstream),
- create a "proxy" calendar in the worker that is not inheriting from ExtendedCalendar (and thus from MemoryCalendar). But ith will have the SQLite backend, and will simply pass data back and forth between the backend and the manager. It will call the storage `load()` methods as now, but instead of storing the data itself, it would signal the manager that new data are available for ownership transfer. On save, the same, the manager would send a cloned version of the data to be saved, and the SQLite backend will use the data but the calendar proxy will not store them.

What do you think of the idea ? Do you see any flaw in the logic ? sailfishos/mkcal#23 is a step in the direction of this proxy calendar, by disconnecting the SQLite backend from ExtendedCalendar. Even if the proxy idea is a flop, I think the mKCal rearrangement would be beneficial to code clarity anyway.

pvuorela:
Can't immediately comment to either direction. Mainly now having generic concerns on adjustment such as:
- Would it effectively increase or reduce complexity and how much.
- How much do we in practice now suffer from the redundant event instances in memory. Would expect heavier calendar usage needed to make difference.
- Pondering how would this play out with other changes we'd want to make. Namely the ability to have one event existing in multiple notebooks is something I'd like to get some day. Now we can't allow moving between notebooks even because that break the deleted event detection.

But the mkcal change so far looked nice (as commented there too)

PR #33, opened 2022-03-18
Use storage observer to do additional action on database changes

dcaliste:
Following the introduction of `::storageUpdated()` from #28, one can now also use the callback method on database modification to actually perform complementary changes like purging deleted local events, or send invitations / updates / cancellations.

This avoids to store events marked as deleted and then try to see if they have actually been deleted to send cancellations fr instance. We're closer to the actual changes that happened in the database.

PS : the API of the service handler is slightly overly redundant : just the storage and the event would be enough. The calendar can be deduced from the storage with `storage->calendar()` and the notebook can also be deduced from the storage and the event with `storage->notebook(storage->calendar()->notebook(event))`.

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -104,7 +104,22 @@ void CalendarWorker::storageUpdated(mKCal::ExtendedStorage *storage,
     Q_UNUSED(storage);
     Q_UNUSED(added);
     Q_UNUSED(modified);
-    Q_UNUSED(deleted);
+
+    for (const KCalendarCore::Incidence::Ptr &event: deleted) {
+        // FIXME: should send response update if deleting an event we have responded to.
+        if (needSendCancellation(event)) {
+            // FIXME: should send cancel only if we own the event
+            event->setStatus(KCalendarCore::Incidence::StatusCanceled);
+            mKCal::ServiceHandler::instance().sendUpdate(event, QString(), mCalendar, mStorage);
+        }
+        // if the event was stored in a local (non-synced) notebook, purge it.
+        const CalendarData::Notebook &notebook = mNotebooks.value(mCalendar->notebook(event));
+        if (notebook.localCalendar
+            && !storage->purgeDeletedIncidences(KCalendarCore::Incidence::List() << event)) {
+            qWarning() << "Failed to purge deleted event" << event->uid()
+                       << "from local calendar" << mCalendar->notebook(event);
+        }
+    }
 
     // No smart handling of known updates at the moment.
     emit storageModifiedSignal();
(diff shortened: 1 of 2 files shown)


dcaliste:
Ah, need a rebase, one moment…

src/calendarworker.cpp
```diff
@@ -102,9 +102,37 @@ void CalendarWorker::storageUpdated(mKCal::ExtendedStorage *storage,
...
+    // Probably those should be combined into a single function on the API, but
+    // until that is done, let's just handle new events as invitations and rest as updates.
+    for (const KCalendarCore::Incidence::Ptr &event: added) {
+        if (event->attendeeCount() > 0) {
+            mKCal::ServiceHandler::instance().sendInvitation(event, QString(), mCalendar, mStorage);
+        }
+    }
+    for (const KCalendarCore::Incidence::Ptr &event: modified) {
+        if (event->attendeeCount() > 0) {
+            mKCal::ServiceHandler::instance().sendUpdate(event, QString(), mCalendar, mStorage);
```
pvuorela, line 117: Hm, this looks like slightly different logic than what it was. Earlier the update seems sent only if attendee list was modified.

pvuorela: Both versions look a bit suspicious. Seems like earlier the update was sent only if the CalendarEventModification attendee list was modified while now it's sent on all the updates. If I'm the organizer then the invitees should be getting updates but if I'm an invitee and get a bright idea of modifying the event I suppose we wouldn't necessarily want to send updates to everyone else?

dcaliste: Indeed, I fully agree with you. And looking at when the updateEventAttendee() was called, it seems to me that just opening the attendee dialog and not modifying anything there (see the conditions to get `dialog.attendeesModified` set to `true` in EventEditPage.qml) will trigger the call to `updateEventAttendees()` in the bindings and subsequently the sending of updates. So it's not very clear when updates are expected to be sent. Only on attendee changes or on any changes, like the date of the meeting ?

I'm not using these features myself, and I've no idea how Outlook for instance is authorizing or not modifications. As you say, I would expect that only the organizer can send updates. But maybe not ? Anyway, I think it's nicer to get all this logic here and discouple it from the routine used to set-up attendee list for the event.

So how do you prefer at Jolla, should we :
- keep the current behaviour, in that case, I remove the update sending from here and put it back in `updateEventAttendees()`,
- or send updates on any change done by the organizer, on attendee or any other characteristic of the event, if so, I'll add the logic here to filter out sending updates if not the organizer,
- or we allow attendees to update the event also and updates are sent in all cases ?

pvuorela: I think it could be best if we start with sending all updates done by the organizer but skip the cases when the user is just an attendee. Like what needSendCancellation() is doing for cancellation (which is kind of a modification too). That method could be possibly renamed as more generic.

dcaliste: Indeed, mirorring the delete case makes sense, I fully agree. I've amended the last commit to reflect this.

pvuorela (approved):

PR #34, opened 2022-03-28
Improve waiting

Tomin1:
I'm not sure if this will fix the issue for good but I think this is a good improvement.

Use `QTRY_COMPARE` instead of comparing in a loop waiting a little bit at a time and then doing `QCOMPARE`. This is able to tell if there is need to wait just a little bit longer for the test to pass. Also since the earlier timeout was three seconds its now five.

That last loop was strange, it was basically a three second delay. I made it to use `QTRY_VERIFY` instead. I also changed the previous wait to use similar non-racy method of waiting.

Also use range-based loops in the test and improve .gitignore file.

### DIFF
--- a/tests/tst_calendarmanager/tst_calendarmanager.cpp
+++ b/tests/tst_calendarmanager/tst_calendarmanager.cpp
@@ -532,13 +532,7 @@ void tst_CalendarManager::test_notebookApi()
     QSignalSpy notebookSpy(mManager, SIGNAL(notebooksChanged(QList<CalendarData::Notebook>)));
 
     // Wait for the manager to open the calendar database, etc
-    for (int i = 0; i < 30; i++) {
-        if (notebookSpy.count() > 0)
-            break;
-
-        QTest::qWait(100);
-    }
-    QCOMPARE(notebookSpy.count(), 1);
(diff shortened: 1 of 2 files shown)


tests/tst_calendarmanager/tst_calendarmanager.cpp
```diff
@@ -592,13 +566,9 @@ void tst_CalendarManager::test_notebookApi()
...
 
+    notebookSpy.clear();
     mManager->setDefaultNotebook(mAddedNotebooks.last()->uid());
-    for (int i = 0; i < 30; i++) {
-        if (notebookSpy.count() > 4)
-            break;
-
-        QTest::qWait(100);
-    }
+    QVERIFY(notebookSpy.wait());
```
pvuorela, line 571: Wouldn't this be racy, in theory at least? Signal could come before wait() is started. QTRY_COMPARE() would handle that?

Tomin1: Yes. I improved it but I'm not sure if it's perfect yet. At least it doesn't have that same issue now.

pvuorela (approved):

PR #35, opened 2022-03-29
[nemo-qml-plugin-calendar] Fix empty organizer.

dcaliste:
Fix a regression introduced in ca935edb where
an organizer is always added to the attendee list
even when empty.


pvuorela (approved):
Makes sense.

PR #36, opened 2022-04-12
[nemo-qml-plugin-calendar] Move away from Notebook::isDefault().

dcaliste:
In relation to sailfishos/mkcal#29 this is deprecating the use of Notebook::isDefault() and replacing it with something related to ExtendedStorage::defaultNotebook() only. @pvuorela, when you have time, thank you in advance.

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -774,7 +774,8 @@ void CalendarWorker::loadNotebooks()
         notebook.uid = mkNotebook->uid();
         notebook.description = mkNotebook->description();
         notebook.emailAddress = mKCal::ServiceHandler::instance().emailAddress(mkNotebook, mStorage);
-        notebook.isDefault = mkNotebook->isDefault();
+        notebook.isDefault = mStorage->defaultNotebook()
+                && (mkNotebook->uid() == mStorage->defaultNotebook()->uid());
         notebook.readOnly = mkNotebook->isReadOnly();
         notebook.localCalendar = mkNotebook->isMaster()
                 && !mkNotebook->isShared()


src/calendarworker.cpp
```diff
@@ -774,7 +774,7 @@ void CalendarWorker::loadNotebooks()
         notebook.uid = mkNotebook->uid();
         notebook.description = mkNotebook->description();
         notebook.emailAddress = mKCal::ServiceHandler::instance().emailAddress(mkNotebook, mStorage);
-        notebook.isDefault = mkNotebook->isDefault();
+        notebook.isDefault = mkNotebook == mStorage->defaultNotebook();
```
pvuorela, line 777: One thing I realized is that this is now pointer comparison, expecting same instances. While that probably works it feels like fragile expectation. * before both operands could compare the actual data but would need some null protection on default notebook side.

Or maybe better could be comparing notebook uids? Currently would get a bit tedious because of needing the null check but maybe ExtendedStorage could have some getter like defaultNotebookUid()?

Related to these, here the Notebook::Ptr type has some feel of cargo cult on kf5-calendarcore api. Thinking would it be better or how easily we could be migrating to const pointer parameters or new copy on return values.

dcaliste: > Or maybe better could be comparing notebook uids?

I fully agree actually, and when rebasing sailfishos/mkcal#26, I settled down to the same conclusion and use there also UIDs to notify of the default notebook.

> here the Notebook::Ptr type has some feel of cargo cult on kf5-calendarcore api. Thinking would it be better or how easily we could be migrating to const pointer parameters or new copy on return values.

I can think of yet another API rework in mKCal to remove the ::Ptr stuff notebooks indeed.

pvuorela: The mkcal PR 29 seemed ok now but to get that in need first some change here. While this is fragile it should do its job, though fetching and comparing the uid should be also ok with the existing APIs. Perhaps could get something on these in first and adjust the ::Ptr side separately. What do you think?

dcaliste: I've updated the patch to use UIDs.

I'm going to think in a separated PR in mKCal how we can change the Notebook API not to use Ptr.

pvuorela: Thanks

PR #40, opened 2023-01-02
[nemo-qml-plugin-calendar] Add an EXDATE to a recurring event when de…

dcaliste:
…leting an exception.

@pvuorela, following the discussion in sailfishos/mkcal#38 here is a PR adding an exdate to a parent event when deleting an exception.

Still WIP, I need to test it.

### DIFF
--- a/src/calendareventquery.cpp
+++ b/src/calendareventquery.cpp
@@ -39,7 +39,7 @@
 #include <QDebug>
 
 CalendarEventQuery::CalendarEventQuery()
-    : mIsComplete(true), mOccurrence(0), mAttendeesCached(false), mEventError(false)
+    : mIsComplete(true), mOccurrence(0), mAttendeesCached(false), mEventError(false), mUpdateOccurrence(false)
 {
     connect(CalendarManager::instance(), SIGNAL(dataUpdated()), this, SLOT(refresh()));
     connect(CalendarManager::instance(), SIGNAL(storageModified()), this, SLOT(refresh()));
@@ -114,6 +114,7 @@ void CalendarEventQuery::setStartTime(const QDateTime &t)
     if (t == mStartTime)
         return;
     mStartTime = t;
+    mUpdateOccurrence = true;
     emit startTimeChanged();
 
     refresh();
@@ -167,7 +168,7 @@ void CalendarEventQuery::doRefresh(CalendarData::Event event, bool eventError)
     if (event.isValid() && (event.uniqueId != mUid || event.recurrenceId != mRecurrenceId))
         return;
 
-    bool updateOccurrence = false;
+    bool updateOccurrence = mUpdateOccurrence;
     bool signalEventChanged = false;
 
     if (event.uniqueId != mEvent.uniqueId || event.recurrenceId != mEvent.recurrenceId) {
@@ -197,6 +198,7 @@ void CalendarEventQuery::doRefresh(CalendarData::Event event, bool eventError)
                 mOccurrence->setParent(this);
             }
         }
+        mUpdateOccurrence = false;
         emit occurrenceChanged();
     }
 
--- a/src/calendareventquery.h
+++ b/src/calendareventquery.h
@@ -161,6 +161,7 @@ private:
     CalendarEventOccurrence *mOccurrence;
     bool mAttendeesCached;
     bool mEventError;
+    bool mUpdateOccurrence;
     QList<CalendarData::Attendee> mAttendees;
 };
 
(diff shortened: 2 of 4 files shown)


dcaliste:
@pvuorela, I've added two commits :
1. one suppressing warnings when updating attendees. There is a strange behaviour in KCalendarCore when calling `setAttendees()`, because it is calling internally `update()` twice in a row without calling `updated()` in between. We can circumvent this bug by declaring a general update with `startUpdate()` in the routine. It makes sense anyway since we're doing several changes.
2. one modifying the event query behaviour to actually emit the `occurenceChanged()` signal when we only change the startTime of the query.

I've added a test ensuring that deleting an exception is adding an EXDATE.

pvuorela (approved):
Fixes the issue, unit tests pass and the change looks good. Approving, thanks.

For one corner was pondering if recurring event occurrence is modified, the exception is made recurring, and then some occurrence of the repeating exception is modified and deleted. The exDateTime should possibly be in the repeating exception occurrence instead of the original parent?

But that's not allowed now by the UI and not even sure how it should go by the .ics spec, so that shouldn't prevent merging this simple fix.

And doesn't necessarily even make sense to allow doing such events in the UI, but for one case pondering would it be nice to have edit option on occurrences like "Edit this and later events", meaning that the original event repeat should end at that point and the replacement to take over from that.

(And similarly occurrence removal should have option to delete occurrence and later ones, i.e. add end time to the repeat)

dcaliste:
> For one corner was pondering if recurring event occurrence is modified, the exception is made recurring, and then some occurrence of the repeating exception is modified and deleted. The exDateTime should possibly be in the repeating exception occurrence instead of the original parent?

In theory yes, but I think that recurring exceptions are not handled at the moment. I may point out this snippet from `KCalendarCore::Calendar::createException()` :
```
    // Recurring exceptions are not support for now
    newInc->clearRecurrence();
```

> would it be nice to have edit option on occurrences like "Edit this and later events", meaning that the original event repeat should end at that point and the replacement to take over from that.

Indeed, and that is supported in `KCalendarCore::OccurrenceIterator` class that we are already using to populate the calendar. I thought about it already, but didn't much further, particularly in the direction of UI changes for this. Maybe I should restart my idea of a page to handle recurring occurrences and exceptions, coming with a third button on the choosing page when trying to modify a recurring event. Besides the already existing "modify this occurrence" and "modify the recurring event", I would like to add a "manage recurrence" one that would open a page listing exceptions (in terms of incidence with a recurring-id, so they can be amended), exdates (so they can be cancelled) and rdates (so they can be individually added). I started a discussion in the forum a while ago, but went sidetracked : https://forum.sailfishos.org/t/changing-exceptions-of-a-recurring-event/3698.

pvuorela:
I'm a bit afraid if such exception managing UI would got techy. Don't think I've ever seen such elsewhere.

> Indeed, and that is supported in KCalendarCore::OccurrenceIterator class

But how would this pan out if recurring exceptions are not supported for now? If "this and later" is edited, it should create a separate instance, supposedly with a recurrence id?

The "delete this and later ones" is possibly the least problematic to implement if it only requires setting end time for recurrence. Would be nice, it is strange that if one wants to get rid of obsolete recurring event the whole thing needs to be deleted even for the past. (Technically could edit the edit manually and set the end time for recurrence, but who has time for that :))

dcaliste:
> Don't think I've ever seen such elsewhere.

Why not being innovative !

> If "this and later" is edited, it should create a separate instance, supposedly with a recurrence id?

It's handled in the `KCalendarCore::OccurrenceIterator`. The iterator is returning an occurrence, meaning a start time and an end time, with an associated incidence (to take the summary from for instance). In the general case, the start time is the one of the incidence. For recurring events, the start time is the one of the occurrence and the incidence is the parent. For exceptions, the start time is the one of the exception and the incidence is the exception. And for case of exception noted `THIS_AND_FUTURE`, the start time is the occurrence one, but the incidence is the exception setting the this and future and all later occurrence use this exception as incidence and not the parent anymore.

> The "delete this and later ones" is possibly the least problematic to implement if it only requires setting end time for recurrence. Would be nice, it is strange that if one wants to get rid of obsolete recurring event the whole thing needs to be deleted even for the past. (Technically could edit the edit manually and set the end time for recurrence, but who has time for that :))

Indeed, one should modify the recurring event to modify the end the recurrence. But having in UI a button to "delete this and future" would be quicker (and it fits nicely beside the existing ones "delete the occurrence" or "delete the series"). The actual implementation being of course to tweak the end time of the recurring parent.

PR #42, opened 2023-01-23
Update plugin

beroset:
In working with this plugin on AsteroidOS, I found that I had to make some changes to allow it to compile and run cleanly with gcc 11.

### DIFF
--- a/src/calendarevent.cpp
+++ b/src/calendarevent.cpp
@@ -130,7 +130,11 @@ CalendarEvent::Recur CalendarEvent::recur() const
 
 QDateTime CalendarEvent::recurEndDate() const
 {
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+    return mData->recurEndDate.endOfDay();
+#else
     return QDateTime(mData->recurEndDate);
+#endif
 }
 
 bool CalendarEvent::hasRecurEndDate() const
--- a/src/calendarimportmodel.cpp
+++ b/src/calendarimportmodel.cpp
@@ -158,7 +158,7 @@ QVariant CalendarImportModel::data(const QModelIndex &index, int role) const
 
 bool CalendarImportModel::importToNotebook(const QString &notebookUid, bool discardInvitation) const
 {
-    for (const KCalendarCore::Incidence::Ptr &incidence : mEventList) {
+    for (const KCalendarCore::Event::Ptr& incidence : mEventList) {
         const KCalendarCore::Incidence::Ptr old =
             mStorage->calendar()->incidence(incidence->uid(), incidence->recurrenceId());
         if (old) {
--- a/src/calendarmanager.cpp
+++ b/src/calendarmanager.cpp
@@ -356,7 +356,7 @@ QList<CalendarData::Range> CalendarManager::addRanges(const QList<CalendarData::
     QList<CalendarData::Range> sortedRanges;
     sortedRanges.append(oldRanges);
     sortedRanges.append(newRanges);
-    qSort(sortedRanges.begin(), sortedRanges.end(), range_lessThan);
+    std::sort(sortedRanges.begin(), sortedRanges.end(), range_lessThan);
 
     // combine
     QList<CalendarData::Range> combinedRanges;
--- a/src/calendarmanager.cpp
+++ b/src/calendarmanager.cpp
@@ -394,12 +394,17 @@ void CalendarManager::updateAgendaModel(CalendarAgendaModel *model)
                 continue;
             }
 
-            const QDateTime startDt(model->startDate()); // To be replaced later by start.startOfDay()
-            const QDateTime endDt(model->endDate()); // To be replaced later by start.startOfDay()
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+            const QDateTime startDt(model->startDate().startOfDay());
+            const QDateTime endDt(model->endDate().endOfDay());
+#else
+            const QDateTime startDt(model->startDate());
+            const QDateTime endDt(QDateTime(model->endDate()).addDays(1));
+#endif
             // on all day events the end time is inclusive, otherwise not
             if ((eo.eventAllDay && eo.startTime.date() <= model->endDate()
                  && eo.endTime.date() >= model->startDate())
-                || (!eo.eventAllDay && eo.startTime < endDt.addDays(1)
+                || (!eo.eventAllDay && eo.startTime < endDt
                     && eo.endTime >= startDt)) {
                 filtered.append(new CalendarEventOccurrence(eo.eventUid, eo.recurrenceId,
                                                             eo.startTime, eo.endTime));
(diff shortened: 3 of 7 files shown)


src/calendarevent.cpp
```diff
@@ -130,7 +130,7 @@ CalendarEvent::Recur CalendarEvent::recur() const
 
 QDateTime CalendarEvent::recurEndDate() const
 {
-    return QDateTime(mData->recurEndDate);
+    return QDateTime(mData->recurEndDate.endOfDay());
```
dcaliste, line 133: This and `.startOfDay()` were introduced later than Qt5.6. SailfishOS is still locked on 5.6 and thus all these changes must be inside `#if` macros on Qt version : /

src/calendarimportmodel.cpp
```diff
@@ -158,7 +158,7 @@ QVariant CalendarImportModel::data(const QModelIndex &index, int role) const
 
 bool CalendarImportModel::importToNotebook(const QString &notebookUid, bool discardInvitation) const
 {
-    for (const KCalendarCore::Incidence::Ptr &incidence : mEventList) {
+    for (const QSharedPointer<KCalendarCore::Event>& incidence : mEventList) {
```
dcaliste, line 161: It can be written as `KCalendarCore::Event::Ptr&` since `mEventList` is a `KCalendarCore::Event::List` (which is internally a `QVector<KCalendarCore::Event::Ptr>`. It's more idiomatic than writing the `QSharedPointer<>`.

dcaliste (changes requested):
Thanks for the changes and modernising the base code. If you needed to add some compilation flags to get the compilation warnings, you may also add them in the `rpm/nemo-qml-plugin-calendar-qt5.spec` or in the `{src,tests}.pro` files depending.

src/calendarevent.cpp
pvuorela: Yea, this. Reverting the QDateTime commit this built fine, though, so others are at least fine on the API availability sense.

beroset: Good point!  I've wrapped each instance of this kind of change within a Qt version check like this:
```
#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
   // new version
#else
   // original version
#endif
```

src/calendarimportmodel.cpp
beroset: Good suggestions!  Done.

src/calendarworker.cpp
```diff
@@ -584,8 +584,13 @@ CalendarWorker::eventOccurrences(const QList<CalendarData::Range> &ranges) const
     const QStringList excluded = excludedNotebooks();
     QHash<QString, CalendarData::EventOccurrence> filtered;
     for (const CalendarData::Range range : ranges) {
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+        KCalendarCore::OccurrenceIterator it(*mCalendar, QDateTime(range.first.addDays(-1).startOfDay()),
```
dcaliste, line 588: I think you can remove the `QDateTime()` here and in the next line since `.startOfDay()` and `.endOfDay()` is already returning a QDateTime object.

src/calendarevent.cpp
```diff
@@ -130,7 +130,11 @@ CalendarEvent::Recur CalendarEvent::recur() const
 
 QDateTime CalendarEvent::recurEndDate() const
 {
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+    return QDateTime(mData->recurEndDate.endOfDay());
```
dcaliste, line 134: Here also, you can remove the `QDateTime()` since `.endOfDay()` is returning a QDateTime object.

tests/tst_calendarevent/tst_calendarevent.cpp
```diff
@@ -154,7 +154,11 @@ void tst_CalendarEvent::modSetters()
     QDateTime recurEnd = QDateTime::currentDateTime().addDays(100);
     eventMod->setRecurEndDate(recurEnd);
     QCOMPARE(recurEndSpy.count(), 1);
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+    QCOMPARE(eventMod->recurEndDate(), QDateTime(recurEnd.date().startOfDay())); // day precision
```
dcaliste, line 158: Here also, remove `QDateTime()` and later on in the tests.

dcaliste (changes requested):
Thanks for the update. Regarding the actual changes, I've got some remarks.

src/calendarmanager.cpp
```diff
@@ -394,8 +394,13 @@ void CalendarManager::updateAgendaModel(CalendarAgendaModel *model)
                 continue;
             }
 
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+            const QDateTime startDt(model->startDate().startOfDay());
+            const QDateTime endDt(model->endDate().endOfDay());
```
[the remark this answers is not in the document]
beroset: Good catch!  Thank you!

dcaliste (approved):
Thanks for the work, it looks nice to me, besides a minor comment. Already approving.

src/calendarmanager.cpp
```diff
@@ -394,12 +394,22 @@ void CalendarManager::updateAgendaModel(CalendarAgendaModel *model)
...
+#else
             const QDateTime startDt(model->startDate()); // To be replaced later by start.startOfDay()
             const QDateTime endDt(model->endDate()); // To be replaced later by start.startOfDay()
```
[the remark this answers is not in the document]
beroset: No apology needed; that's a good comment.  Done!

dcaliste:
Thanks for the update, looks good to me. I'll run the tests tomorrow to be sure.

src/calendarmanager.cpp
```diff
@@ -394,12 +394,17 @@ void CalendarManager::updateAgendaModel(CalendarAgendaModel *model)
...
             }
 
-            const QDateTime startDt(model->startDate()); // To be replaced later by start.startOfDay()
-            const QDateTime endDt(model->endDate()); // To be replaced later by start.startOfDay()
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+            const QDateTime startDt(model->startDate().startOfDay());
+            const QDateTime endDt(model->endDate().endOfDay());
+#else
+            const QDateTime startDt(model->startDate());
+            const QDateTime endDt(model->endDate()).addDays(1);
```
dcaliste, line 402: Ah, actually, this doesn't compile, it should be written like `const QDateTime endDt(QDateTime(model->endDate()).addDays(1));`

beroset: I was confused since I compiled and tested it, but then I realized I was only testing the parts with the newer version of Qt.  My apologies!

pvuorela:
Running the tests myself they pass fine. The Qt <5.14 changes don't either seem too big.

Would be fine for me already, but waiting a moment if @dcaliste still had some comments.

src/calendarmanager.cpp
dcaliste: No problem.

dcaliste:
Thanks @pvuorela for testing also. I guess the failing ones come from my setup (I'll investigate later, since it's not related to the changes here). It's all good to me.

pvuorela (approved):
Alright, let's get this in. Thanks for the PR!

PR #43, opened 2023-02-13
[nemo-qml-plugin-calendar] Add a thisAndFuture attribute to EventModi…

dcaliste:
…fication objects.

This attribute is used for EventModification objects created to modify an occurrence. It is saved to the database and allow to apply modifications to a recurring event for all future occurrences of a given date.

@pvuorela, it requires sailfishos/mkcal#46 to be built.


dcaliste:
Indeed, I still need to polish the UI part using it ; )

dcaliste:
I've added two commits on top adding support to delete occurrences later than a this and future exception or to modify them. It requires to add a shift to get the proper recurrenceId, suitable for the parent, from the startTime of the occurrence, due to possible change in dtStart with the this and future exception used as a base.

PR #44, opened 2023-02-21
[nemo-qml-plugin-calendar] Since mKCal 0.7.6, load is loading the ful…

dcaliste:
…l series.

Don't use loadSeries() and load(uid, recid) from mKCal, since load(uid) is now always loading the parent and its exceptions.

@pvuorela, this is based on sailfishos/mkcal#47. It allow some simplifications in the code. Notice also that `mCalendar->deleteEventInstances()` is actually called by KCalendarCore internally when deleting a recurring event (see KCalendarCore/src/memorycalendar.cpp#230). So no need to call it here also.

### DIFF
--- a/rpm/nemo-qml-plugin-calendar-qt5.spec
+++ b/rpm/nemo-qml-plugin-calendar-qt5.spec
@@ -1,7 +1,7 @@
 Name:       nemo-qml-plugin-calendar-qt5
 
 Summary:    Calendar plugin for Nemo Mobile
-Version:    0.6.48
+Version:    0.6.53
 Release:    1
 License:    BSD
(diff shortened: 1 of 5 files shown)


pvuorela:
Scratching my head a bit on what are the simplifications allowed by the change, as loadSeries() switched to new load(uid) should effectively be more or less the same thing.

Anyhow, would be good to mention about simplifications and deleteEventInstances sort of details in the commit message as that's what's going to be read if ever getting back to why a commit was done.

pvuorela:
I guess this is fine, but wouldn't hurt having a bit more explanation on the commit message.

pvuorela (approved):

PR #45, opened 2023-02-27
[nemo-qml-plugin-calendar] Add a search model.

dcaliste:
Add a search function to CalendarManager. It is based on ExtendedStorage::search() method. The worker is loading all matching incidences and retrieving the corresponding instance identifiers. The later are
passed back to the manager via a new searchResults signal. While the former are sent back viw the
existing dataLoaded signal.

Also add a model listing occurrence of a search
query. This model is based on a EventListModel
where the identifiers are given as the result of
a search call.

@pvuorela, this is based on sailfishos/mkcal#49 and aims at providing a list model for search results. It is conveniently based on evntlistmodel, since such model already provide event listing from instance identifiers. I still keep this as WIP since I would like to add a loading property to the eventlistmodel. But having it public already should help to see how the mkcal PR is working in reality.

### DIFF
--- a/src/calendareventlistmodel.h
+++ b/src/calendareventlistmodel.h
@@ -44,6 +44,7 @@ class CalendarEventListModel : public QAbstractListModel, public QQmlParserStatu
     Q_PROPERTY(int count READ count NOTIFY countChanged)
     Q_PROPERTY(QStringList identifiers READ identifiers WRITE setIdentifiers NOTIFY identifiersChanged)
     Q_PROPERTY(QStringList missingItems READ missingItems NOTIFY missingItemsChanged)
+    Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
 
 public:
     enum EventListRoles {
@@ -63,6 +64,8 @@ public:
 
     QStringList missingItems() const;
 
+    bool loading() const;
+
     int rowCount(const QModelIndex &index) const;
     QVariant data(const QModelIndex &index, int role) const;
 
@@ -73,6 +76,7 @@ signals:
     void countChanged();
     void identifiersChanged();
     void missingItemsChanged();
+    virtual void loadingChanged();
 
 protected:
     virtual QHash<int, QByteArray> roleNames() const;
--- a/src/calendarsearchmodel.cpp
+++ b/src/calendarsearchmodel.cpp
@@ -54,10 +54,23 @@ void CalendarSearchModel::setSearchString(const QString &searchString)
     setIdentifiers(QStringList());
     if (!mSearchString.isEmpty()) {
         CalendarManager::instance()->search(this);
+        emit loadingChanged();
     }
 }
 
 QString CalendarSearchModel::searchString() const
 {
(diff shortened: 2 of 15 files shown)


dcaliste:
I've added a filter on calendar visibility in the worker thread to send to the manager only visible events.

dcaliste:
I've also added a loading attribute to the EventListModel, so one can display a busy indicator while the model is searching.

src/calendarsearchmodel.cpp
```diff
@@ -0,0 +1,108 @@
...
+    if (!data.isValid())
+        return QVariant();
+
+    CalendarEventOccurrence *occurrence = qobject_cast<CalendarEventOccurrence*>(data.value<QObject*>());
+    if (!occurrence)
+        return QVariant();
+
+    switch (role) {
+    case SectionRole:
+        return QVariant(occurrence->startTime().date().year());
```
pvuorela, line 71: Section is what this can get used for but the role could be more like "year".

src/calendareventlistmodel.h
```diff
@@ -44,12 +44,14 @@ class CalendarEventListModel : public QAbstractListModel, public QQmlParserStatu
...
     Q_PROPERTY(QStringList missingItems READ missingItems NOTIFY missingItemsChanged)
+    Q_PROPERTY(bool loading READ loading NOTIFY loadingChanged)
 
 public:
     enum EventListRoles {
         EventObjectRole = Qt::UserRole,
         OccurrenceObjectRole,
-        IdentifierRole
+        IdentifierRole,
+        ChildRole // Used by classes inheriting CalendarEventListModel
```
pvuorela, line 54: Think more often the name of the last item in enum is having the word Last in rather than Child. Alternatively could just do something like EventObjectRole +100 on the inherited classes first item to define some number range which should suffice here.

dcaliste: Ok, I went for `LastRole`. Thanks for the suggestion.

src/calendarsearchmodel.cpp
dcaliste: You're right. The role is now called `YearRole` with a string represention of `year`.

dcaliste:
I've added a commit adding a `limit` property to the search model. It has no sense in my opinion to put it in the EventListModel since the later is not sorted while the former is.

dcaliste:
Rebased on latest master changes.

src/calendarworker.cpp
```diff
@@ -752,6 +752,39 @@ CalendarData::Event CalendarWorker::createEventStruct(const KCalendarCore::Event
     return event;
 }
 
+void CalendarWorker::search(const QString &pattern, int limit)
```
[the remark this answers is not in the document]
dcaliste: Right, I rename the property to `searchString` and change it everywhere accordingly.

dcaliste:
I've removed the WIP notice.

There may be some modification still to properly handle next occurrence of recurring events, but it can be done in a later PR.

PR #46, opened 2023-03-03
[nemo-calendar] Remove deprecated use of Incidence::setHasGeo(). Fixes JB#60292

pvuorela:
No-op if bool parameter is true and otherwise there area lready calls to setGeoLatitude/Longitude.

@dcaliste

### DIFF
--- a/tools/icalconverter/main.cpp
+++ b/tools/icalconverter/main.cpp
@@ -330,7 +330,6 @@ namespace CalendarImportExport {
             COPY_IF_NOT_EQUAL(dest, src, description(), setDescription);
             COPY_IF_NOT_EQUAL(dest, src, geoLatitude(), setGeoLatitude);
             COPY_IF_NOT_EQUAL(dest, src, geoLongitude(), setGeoLongitude);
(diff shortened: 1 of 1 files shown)


dcaliste:
Yes, no need to setHasGeo(), by construction with the previous setGeoLat/Lon, the src.hasGeo and dest.hasGeo must be equal.

Good to me.

Tomin1 (approved):
> No-op if bool parameter is true and otherwise there _area lready_ calls to
setGeoLatitude/Longitude.

Please check the commit message for typos.

dcaliste:
Ideally, one should do something like:
```
 dest->startUpdates();
 *dest.staticCast<KCalendarCore::IncidenceBase>() = *src.staticCast<KCalendarCore::IncidenceBase>();
 dest->endUpdates();
```
and maybe reset some properties that we may not want to copy (if any).

pvuorela:
> Ideally, one should do something like

Yea, could be. Not sure immediately why there is such an explicit property handling, but can be considered then separately some day.

PR #47, opened 2023-03-08
[nemo-qml-plugin-calendar] Reset calendar on timezone change on device.

dcaliste:
Create a hard dependency on Timed to get the timezone change info.

@pvuorela : this is allowing to change the timezone in the system setting on device and the calendar still displays events at the right date and time. If the event is floating, no change is done. If the event is given by UTC or in a timezone, the display time is shifted accordingly. If it requires a date change, the event is moved and agenda models are updated accordingly. Maybe you have a long standing JB id available for this ? ;-)

The `emitStorageModified()` may seem a bit overkill, but the caching mechanism in mKCal will make it much less resource intensive. The ranges won't be reloaded for instance, just the occurrences will be recomputed (which is desired). All the events by UID will be reloaded though. I may add a safeguard in mKCal for this also : there is no necessity to reload an event that we already have in the calendar. Either it's not modified and it's a possibly blocking call to read the DB to get the same data, or the DB has been modified externally and anyway the calendar has been flushed already and emptied from any events.

### DIFF
--- a/src/calendaragendamodel.cpp
+++ b/src/calendaragendamodel.cpp
@@ -44,6 +44,8 @@ CalendarAgendaModel::CalendarAgendaModel(QObject *parent)
 {
     connect(CalendarManager::instance(), SIGNAL(storageModified()), this, SLOT(refresh()));
     connect(CalendarManager::instance(), SIGNAL(dataUpdated()), this, SLOT(refresh()));
+    connect(CalendarManager::instance(), &CalendarManager::timezoneChanged,
+            this, &CalendarAgendaModel::onTimezoneChanged);
 }
 
 CalendarAgendaModel::~CalendarAgendaModel()
@@ -283,6 +285,18 @@ QVariant CalendarAgendaModel::get(int index, int role) const
     default:
         qWarning() << "CalendarAgendaModel: Unknown role asked";
         return QVariant();
+    }
+}
+
+void CalendarAgendaModel::onTimezoneChanged()
+{
+    QList<CalendarEventOccurrence *>::ConstIterator it;
+    for (it = mEvents.constBegin(); it != mEvents.constEnd(); it++) {
+        // Actually, the date times have not changed, but
+        // their representations in local time (as used in QML)
+        // have changed.
+        (*it)->startTimeChanged();
+        (*it)->endTimeChanged();
     }
 }
 
--- a/src/calendaragendamodel.h
+++ b/src/calendaragendamodel.h
@@ -103,6 +103,7 @@ protected:
 
 private slots:
     void refresh();
+    void onTimezoneChanged();
 
 private:
     QDate mStartDate;
--- a/src/src.pro
+++ b/src/src.pro
@@ -2,7 +2,7 @@ TARGET = nemocalendar
 PLUGIN_IMPORT_PATH = org/nemomobile/calendar
 
 TEMPLATE = lib
-CONFIG += qt plugin hide_symbols
+CONFIG += qt plugin hide_symbols timed-qt5
 
 QT += qml concurrent
 QT -= gui
--- a/lightweight/calendardataservice/calendardataservice.pro
+++ b/lightweight/calendardataservice/calendardataservice.pro
@@ -5,7 +5,7 @@ target.path = /usr/bin
 QT += qml dbus
 QT -= gui
 
-CONFIG += link_pkgconfig
+CONFIG += link_pkgconfig timed-qt5
 PKGCONFIG += KF5CalendarCore libmkcal-qt5 accounts-qt5
 
 HEADERS += \
--- a/rpm/nemo-qml-plugin-calendar-qt5.spec
+++ b/rpm/nemo-qml-plugin-calendar-qt5.spec
@@ -13,6 +13,7 @@ BuildRequires:  pkgconfig(Qt5Concurrent)
 BuildRequires:  pkgconfig(libmkcal-qt5) >= 0.7.6
 BuildRequires:  pkgconfig(KF5CalendarCore)
 BuildRequires:  pkgconfig(accounts-qt5)
+BuildRequires:  pkgconfig(timed-qt5)
 
 %description
 %{summary}.
--- a/src/calendareventlistmodel.cpp
+++ b/src/calendareventlistmodel.cpp
@@ -43,6 +43,8 @@ CalendarEventListModel::CalendarEventListModel(QObject *parent)
             this, &CalendarEventListModel::refresh);
     connect(CalendarManager::instance(), &CalendarManager::dataUpdated,
             this, &CalendarEventListModel::doRefresh);
+    connect(CalendarManager::instance(), &CalendarManager::timezoneChanged,
+            this, &CalendarEventListModel::onTimezoneChanged);
 }
 
 CalendarEventListModel::~CalendarEventListModel()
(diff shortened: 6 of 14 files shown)


dcaliste:
I've added sailfishos/mkcal#55 for this minor optimisation with loading by UIDs.

src/src.pro
```diff
@@ -4,12 +4,12 @@ PLUGIN_IMPORT_PATH = org/nemomobile/calendar
 TEMPLATE = lib
 CONFIG += qt plugin hide_symbols
 
-QT += qml concurrent
+QT += qml concurrent dbus
```
pvuorela, line 7: This looks like timed .pc file shortcoming. If using timed requires that, it should be timed's responsibility to require it. Suppose needing Requires: line there.

dcaliste: Ah yes, added `BuildRequires:  pkgconfig(Qt5DBus)`

pvuorela: That's another thing the timed-devel should handle. If using that needs Qt5DBus, then timed-qt5-devel should have the package dependency.

dcaliste: Ok, so I've created sailfishos/timed#6 and removed this BuildRequires.

I've also changed the pro files not to use the pkgconfig file, but instead I've added the `timed-qt5` as a `CONFIG` in the pro files, since timed is providing mkspecs feature files.

tools/icalconverter/icalconverter.pro
```diff
@@ -1,7 +1,7 @@
 TEMPLATE=app
 TARGET=icalconverter
 QT-=gui
-CONFIG += link_pkgconfig
+CONFIG += link_pkgconfig timed-qt5
```
pvuorela, line 4: Don't suppose this should be needed, timed only in the worker class which isn't used here?

dcaliste: Mmh that's correct. I've removed it. I added by mistake, misunderstanding an error in another directory but printed just after the compilation of this tool…

pvuorela:
> Maybe you have a long standing JB id available for this ? ;-)

Oh yes, JB#42335. Only 5 years old, but the original JB#14935 got at some point replaced with a newer instance :)

dcaliste:
Thanks, I've added the bug id to the commit message.

pvuorela:
Testing with jolla calendar this otherwise seems to update the times but for the event list of the current date they don't update. If I switch to some other day and back, then the times are according to the new timezone.

dcaliste:
The reason of this is in the case for the AgendaModel not to change count, it is not emitting `dataChanged()` for the events that were already in the model. I'm adding such a signal.

I'm wondering though if it will not create a performance penalty because it is emitted all the time for every event in the model that was not removed. Besides, as mentioned in the comment added in the code, it would also solve a possible issue with background sync changing summary or date (while I'm not sure 100% because a background sync will reset the event list in the manager…).

src/calendaragendamodel.cpp
```diff
@@ -197,6 +197,10 @@ void CalendarAgendaModel::doRefresh(QList<CalendarEventOccurrence *> newEvents)
         // Skip matching events
         while (eventsCounter < events.count() && newEventsCounter < newEvents.count() &&
                eventsEqual(newEvents.at(newEventsCounter), events.at(eventsCounter))) {
+            // We don't know if the event has changed or not,
+            // it may have been reloaded with a different time or summary
+            // because of background sync or timezone change.
+            dataChanged(index(eventsCounter), index(eventsCounter));
```
pvuorela, line 203: Hm, the eventsEqual() check above gets my first attention. If it's returning true should they be just the same events or with the same content as earlier? Now the check is somewhere in between when it checks the time and uid, but not the other data. Maybe there could be some possibility for better comparison, or maybe separate uid and the other things.

pvuorela: For another thing, the model here serves data with three roles, like event occurrence and event pointers. If those values don't change, there shouldn't ideally be dataChanged() signalling.

Could be that the timezone change is a special case and the summary and such are already working? Now the start/end time labels seem to be coming from model.occurrence, and marked CONSTANT. Pondering if simplest just to detect the timezone change and consider all the events changed.

dcaliste: Yes, I fully agree. I was planning to expose the time zone change to the manager, so it can emit date time change on manage objects.

dcaliste:
@pvuorela I've revisited the time zone change. The worker is transmitting the time zone change to the manager. Then, all occurrences and event objects are notifying on their start time and end time. The event objects are owned by the manager, so it's simple. The occurrences are owned by their respective model (agenda, event list or event query). The time zone change is thus transmitted by the manager to these models.

I've added a commit on top factorizing code between agenda models and event list models.

There is a missing model with an event occurrence that does not update on time zone change : the import model. Indeed, it is populated with ImportEvent objects that spawn EventOccurrence objects on-demand in the `getNextOccurrence()` method. It is not keeping track of the occurrence. I didn't touch to this code though, because I don't remember much the ownership policy used in QML. Does the QML engine taking ownership on a returned QObject* in a Q_INVOKABLE method ? If yes, then, it's useless to keep a pointer on this new object since it can be killed by QML without notice. If no, this code is broken and leaking. What do you think ? Should we kill the `getNextOccurrence()` method and create a QObject* property called `occurrence` with a proper ownership to the ImportEvent object (and then possibly notify on time zone change for this occurrence also) ?

src/calendaragendamodel.h
```diff
@@ -67,49 +60,39 @@ class CalendarAgendaModel : public QAbstractListModel, public QQmlParserStatus
     Q_DECLARE_FLAGS(FilterModes, FilterMode)
     Q_FLAG(FilterModes)
 
-    explicit CalendarAgendaModel(QObject *parent = 0);
+    CalendarAgendaModel(QObject *parent = 0);
```
pvuorela, line 63: Why remove the explicit?

src/calendaragendamodel.h
```diff
@@ -67,49 +60,39 @@ class CalendarAgendaModel : public QAbstractListModel, public QQmlParserStatus
...
     int filterMode() const;
     void setFilterMode(int mode);
 
     // CalendarAgendaModel takes ownership of the CalendarEventOccurrence objects
     void doRefresh(QList<CalendarEventOccurrence *>);
 
-    int rowCount(const QModelIndex &index) const;
-    QVariant data(const QModelIndex &index, int role) const;
-    Q_INVOKABLE QVariant get(int index, int role) const;
+    Q_INVOKABLE virtual QVariant get(int index, int role) const;
```
pvuorela, line 80: Optionally could be marked override, some other methods too. Doesn't matter much.

src/calendaragendamodel.cpp
```diff
@@ -96,9 +99,6 @@ void CalendarAgendaModel::setEndDate(const QDate &endDate)
 
 void CalendarAgendaModel::refresh()
 {
-    if (!mIsComplete)
-        return;
```
pvuorela, line 100: This doesn't look right. refresh() is called from property setters which can happen before the component is complete, but those should not cause any handling in the intermediate state until all the properties are set.

pvuorela (commented):
The latter commit breaks the agenda model unit test. 

I can understand combing the common code to a base class, but now that it gets used only for two models, the result is one class more and even not counting the header lines, it doesn't end up with less code than earlier. So I'd maybe just skip that now and we could merge the first one. That's looking good to me already.

pvuorela (approved):

dcaliste:
Thanks for merging. Do you have any remark on the import event stuff (see my discussion above) ? Because for the moment, if you have an import dialog opened and you change timezone (ok, very very far fetched case), the event times and dates are not updated.

PR #49, opened 2023-04-01
Include dbus for timed-qt5/qmacro.h

jmlich:
Without dbus module compilation fails with following error:

```
In file included from /usr/include/timed-qt5/wall-declarations.h:32,
                 from calendarworker.h:48,
                 from calendarmanager.cpp:37:
/usr/include/timed-qt5/qmacro.h:29:10: fatal error: QDBusMetaType: No such file or directory
   29 | #include <QDBusMetaType>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -4,7 +4,7 @@ PLUGIN_IMPORT_PATH = org/nemomobile/calendar
 TEMPLATE = lib
 CONFIG += qt plugin hide_symbols timed-qt5
 
-QT += qml concurrent
+QT += qml concurrent dbus
 QT -= gui
 QMAKE_CXXFLAGS += -Werror
 


jmlich:
Sorry, for noise. It seems the timed-qt5.prf is not loaded for some reason

dcaliste:
I think this is solved in timed-devel with https://github.com/sailfishos/timed/pull/6

jmlich:
I should probably add an explanation. In Manjaro the *.prf files are loaded from /usr/lib/qt/mkspecs/features, but timed-qt5.prf is installed into /usr/share/.. and not found during build process. The consequence was that dbus wasn't added at all. 

I need to adjust way how the timed is built in order to put *.prf into folder where OS expects it. I have also noticed that some packages are using /usr/share/qt5. I can probably solve this in build process it self, by defining the position of it as qmake argument. Maybe I will come with some smart solution yet.

dcaliste:
Ah, I see. Thanks for the explanation. I may have made your life easier if I introduced timed as a PKG_CONFIG and not as a CONFIG in src.pro... Sorry :/

I've no idea which location is standard for .prf files. In Debian, they are also under /usr/lib/x86_64-linux-gnu/qt5. I guess you'll have to patch timed to put the .prf file in the expected location.

PR #51, opened 2023-05-11
[nemo-qml-plugin-calendar] Specify the notebook when purging.

dcaliste:
Provides the notebook UID of the event to
purge. Not a necessity at the moment since
event UIDs are unique in the database, but
it will allow to raise the unicity constrain
between notebooks later on.

@pvuorela : it's a follow-up of sailfishos/mkcal#59.


pvuorela (approved):
Nice and simple

PR #54, opened 2023-05-23
Make Event::uniqueId sole identifier for events.

dcaliste:
Remove the recurrenceId from the API, so uniqueId
    is able to uniquely identify an event, or an exception.
    This is done internally by affecting the
    instanceIdentifier to Event::uniqueId.
    
This would allow later to add a notebook UID
    information to this uniqueId and thus being
    able to add multi-notebook support without
    changing once again the API.

@pvuorela , this one is more to open a discussion on the evolution of the nemo-qml-plugin-calendar API, because with incidences that can share the same UID between notebooks, obviously things like Manager::eventObject(uid, recId) or CalendarApi::remove(uid, recurrenceId, startTime) are broken. I can see three ways to solve this:
1. Add a notebookUid argument to such routines, which means an API break for the public routines or additionnal required properties like for CalendarEventQuery.
2. Hide the notebookUid somehow in the uid by setting CalendarData::Event::uniqueId to notebookUid+uid. This is a  bit ugly to hide something in something else but it keeps the API untouched,
3. Assume the change of meaning of CalendarData::Event::uniqueId into something that is truely a unique identifier between events, exceptions and notebooks, and remove the recurrenceId from such API, just keeping this uniqueId that hold everything. Currently, it can simply be the instanceIdentifier() of an incidence, but later, we can extend it easily with the notebookUid.

The PR here investigate the consequences of option 3, namely dropping the recurrenceId argument from all API used to uniquely identify an event. It starts on #53 obviously since this PR is moving in that direction for the internal storage. In term of API changes, it means:
* CalendarEventQuery drops the recurrenceId property and simply uses the uniqueId one to identify an event. The startTime is kept to get a specific occurrence for recurring events.
* CalendarApi::remove() drops the recurrenceId argument.

These API changes introduce changes in three UI applications as far as my grepping capability is concerned:
* jolla-calendar, the largest changes, but quite strait forward (I've a PR ready to show it). It implies though a change in its DBus API for "viewEvent()" function. Once again grepping for it in "/usr/share" on device or in "/usr/lib64/qt5/qml" reveals only one match : the alarms, see later.
* jolla-email, a direct change not to pass the recurrenceId read from the invitation query since the uniqueId is now enough. So very simple.
* jolla-alarm, in `/usr/share/jolla-alarm-ui/pages/CalendarAlarmDialog.qml`. There, the pull up menu has an option to show the alarm in calendar, thus calling the viewEvent DBus function. The recurrenceId can be drop here.

This alarm stuff raise a transition issue: one can store the uniqueId in mKCal when setting the alarm, so it is available in the ringing dialog and one can pass it to viewEvent(). So far so good. But what happens to previously set alarms that were not defined with the uniqueId in mind ? What is good is that they already contain the three information, the notebookUid, the event uid and the recurrenceId. So it can be possible to reconstruct the uniqueId from these three info. But it's a bit ugly…

### DIFF
--- a/lightweight/common/eventdata.cpp
+++ b/lightweight/common/eventdata.cpp
@@ -4,8 +4,7 @@ QDBusArgument &operator<<(QDBusArgument &argument, const EventData &eventData)
 {
     argument.beginStructure();
     argument << eventData.calendarUid
-             << eventData.uniqueId
-             << eventData.recurrenceId
+             << eventData.instanceId
              << eventData.startTime
              << eventData.endTime
              << eventData.allDay
@@ -22,8 +21,7 @@ const QDBusArgument &operator>>(const QDBusArgument &argument, EventData &eventD
 {
     argument.beginStructure();
     argument >> eventData.calendarUid
-             >> eventData.uniqueId
-             >> eventData.recurrenceId
+             >> eventData.instanceId
              >> eventData.startTime
              >> eventData.endTime
              >> eventData.allDay
--- a/src/calendarevent.cpp
+++ b/src/calendarevent.cpp
@@ -157,9 +157,9 @@ QDateTime CalendarEvent::reminderDateTime() const
     return mData->reminderDateTime;
 }
 
-QString CalendarEvent::uniqueId() const
+QString CalendarEvent::instanceId() const
 {
-    return mData->uniqueId;
+    return mData->instanceId;
 }
 
 bool CalendarEvent::readOnly() const
@@ -248,9 +248,9 @@ void CalendarStoredEvent::notebookColorChanged(QString notebookUid)
 
 void CalendarStoredEvent::eventUidChanged(QString oldUid, QString newUid)
 {
-    if (mData->uniqueId == oldUid) {
-        mData->uniqueId = newUid;
-        emit uniqueIdChanged();
+    if (mData->instanceId == oldUid) {
+        mData->instanceId = newUid;
+        emit instanceIdChanged();
         // Event uid changes when the event is moved between notebooks, calendar uid has changed
         emit calendarUidChanged();
     }
@@ -258,7 +258,7 @@ void CalendarStoredEvent::eventUidChanged(QString oldUid, QString newUid)
 
 bool CalendarStoredEvent::sendResponse(int response)
 {
-    if (mManager->sendResponse(mData->uniqueId, mData->recurrenceId, (Response)response)) {
+    if (mManager->sendResponse(mData->instanceId, (Response)response)) {
         mManager->save();
         return true;
     } else {
@@ -268,7 +268,7 @@ bool CalendarStoredEvent::sendResponse(int response)
 
 void CalendarStoredEvent::deleteEvent()
 {
-    mManager->deleteEvent(mData->uniqueId, mData->recurrenceId, QDateTime());
+    mManager->deleteEvent(mData->instanceId, QDateTime());
     mManager->save();
 }
 
@@ -276,13 +276,13 @@ void CalendarStoredEvent::deleteEvent()
 QString CalendarStoredEvent::iCalendar(const QString &prodId) const
 {
     Q_UNUSED(prodId);
-    if (mData->uniqueId.isEmpty()) {
+    if (mData->instanceId.isEmpty()) {
         qWarning() << "Event has no uid, returning empty iCalendar string."
                    << "Save event before calling this function";
         return QString();
     }
 
-    return mManager->convertEventToICalendarSync(mData->uniqueId, prodId);
+    return mManager->convertEventToICalendarSync(mData->instanceId, prodId);
 }
 
 QString CalendarStoredEvent::color() const
@@ -332,5 +332,5 @@ void CalendarStoredEvent::setEvent(const CalendarData::Event *data)
 
 CalendarData::Event CalendarStoredEvent::dissociateSingleOccurrence(const CalendarEventOccurrence *occurrence) const
 {
-    return occurrence ? mManager->dissociateSingleOccurrence(mData->uniqueId, occurrence->startTime()) : CalendarData::Event();
+    return occurrence ? mManager->dissociateSingleOccurrence(mData->instanceId, occurrence->startTime()) : CalendarData::Event();
 }
--- a/src/calendarevent.cpp
+++ b/src/calendarevent.cpp
@@ -36,6 +36,8 @@
 #include <QDateTime>
 #include <QTimeZone>
 
+#include <KCalendarCore/Event>
+
 #include "calendarutils.h"
 #include "calendarmanager.h"
 #include "calendareventoccurrence.h"
@@ -162,6 +164,11 @@ QString CalendarEvent::instanceId() const
     return mData->instanceId;
 }
 
+bool CalendarEvent::isException() const
+{
+    return mData->recurrenceId.isValid();
+}
+
 bool CalendarEvent::readOnly() const
 {
     return mData->readOnly;
@@ -210,20 +217,6 @@ bool CalendarEvent::rsvp() const
 bool CalendarEvent::externalInvitation() const
 {
     return mData->externalInvitation;
-}
-
-QDateTime CalendarEvent::recurrenceId() const
-{
-    return mData->recurrenceId;
-}
-
-QString CalendarEvent::recurrenceIdString() const
-{
-    if (mData->recurrenceId.isValid()) {
-        return CalendarUtils::recurrenceIdToString(mData->recurrenceId);
-    } else {
-        return QString();
-    }
 }
 
 CalendarStoredEvent::CalendarStoredEvent(CalendarManager *manager, const CalendarData::Event *data)
@@ -283,6 +276,17 @@ QString CalendarStoredEvent::iCalendar(const QString &prodId) const
     }
 
     return mManager->convertEventToICalendarSync(mData->instanceId, prodId);
+}
+
+CalendarStoredEvent* CalendarStoredEvent::parent() const
+{
+    if (isException()) {
+        KCalendarCore::Event event;
+        event.setUid(mData->incidenceUid);
+        return mManager->eventObject(event.instanceIdentifier());
+    } else {
+        return nullptr;
+    }
 }
 
 QString CalendarStoredEvent::color() const
--- a/src/calendarevent.cpp
+++ b/src/calendarevent.cpp
@@ -225,8 +225,8 @@ CalendarStoredEvent::CalendarStoredEvent(CalendarManager *manager, const Calenda
 {
     connect(mManager, SIGNAL(notebookColorChanged(QString)),
             this, SLOT(notebookColorChanged(QString)));
-    connect(mManager, SIGNAL(eventUidChanged(QString,QString)),
-            this, SLOT(eventUidChanged(QString,QString)));
+    connect(mManager, &CalendarManager::instanceIdChanged,
+            this, &CalendarStoredEvent::instanceIdNotified);
 }
 
 CalendarStoredEvent::~CalendarStoredEvent()
@@ -239,13 +239,15 @@ void CalendarStoredEvent::notebookColorChanged(QString notebookUid)
         emit colorChanged();
 }
 
-void CalendarStoredEvent::eventUidChanged(QString oldUid, QString newUid)
+void CalendarStoredEvent::instanceIdNotified(QString oldId, QString newId, QString notebookUid)
 {
-    if (mData->instanceId == oldUid) {
-        mData->instanceId = newUid;
+    if (mData->instanceId == oldId) {
+        mData->instanceId = newId;
         emit instanceIdChanged();
         // Event uid changes when the event is moved between notebooks, calendar uid has changed
+        mData->calendarUid = notebookUid;
         emit calendarUidChanged();
+        emit colorChanged();
     }
 }
 
--- a/src/calendarevent.h
+++ b/src/calendarevent.h
@@ -225,7 +225,7 @@ signals:
 
 private slots:
     void notebookColorChanged(QString notebookUid);
-    void eventUidChanged(QString oldUid, QString newUid);
+    void instanceIdNotified(QString oldId, QString newId, QString notebookUid);
 
 private:
     CalendarManager *mManager;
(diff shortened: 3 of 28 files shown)


pvuorela:
Overall I think it's nice having a single really unique identifier for the events. Just thinking should it be called something a bit different to reduce confusion with uid of an event, e.g. instanceId or something like that. The API and its usage need to be touched anyway and naming would be hopefully just mechanical search & replace.

No immediate solution for the alarms. Guess there could be some backward compatibility support, detecting how the alarm properties were set up. Or then keep supporting showing events based on uid+recid. In theory multiple notebooks could have the same event but out of sync where it would matter what notebook to use for showing, but then again it's strange even if they are in sync and there are alarms for each notebook separately. Or maybe that's what the user has ordered if the same thing is in multiple places.

dcaliste:
Rebased. It seems I had included in this PR the initial commit of the manager fix on model destruction…

src/calendarmanager.h
```diff
@@ -61,22 +61,22 @@ class CalendarManager : public QObject
...
     // Synchronous DB thread access
-    QString convertEventToICalendarSync(const QString &uid, const QString &prodId);
+    QString convertEventToICalendarSync(const QString &instanceId, const QString &prodId);
 
     // Event
-    CalendarData::Event getEvent(const QString& instanceIdentifier, bool *loaded = nullptr) const;
-    CalendarData::Event dissociateSingleOccurrence(const QString &eventUid, const QDateTime &recurrenceId) const;
-    bool sendResponse(const QString &uid, const QDateTime &recurrenceId, CalendarEvent::Response response);
+    CalendarData::Event getEvent(const QString& instanceId, bool *loaded = nullptr) const;
+    CalendarData::Event dissociateSingleOccurrence(const QString &instanceId, const QDateTime &recurrenceId) const;
```
pvuorela, line 78: This should still be unique id since it's paired with recurrence id?

src/calendarutils.cpp
```diff
@@ -53,7 +53,8 @@ CalendarData::Event::Event(const KCalendarCore::Event &event)
     , startTime(event.dtStart())
     , endTime(event.dtEnd())
     , allDay(event.allDay())
-    , uniqueId(event.uid())
+    , uniqueId(event.instanceIdentifier())
+    , incidenceId(event.uid())
```
pvuorela, line 57: This seems a bit confusing naming when uniqueId is not the same thing as uid.  First I thought incidenceId was the instance identifier.

pvuorela (commented):
Getting still a bit confused with the names when there's uid / uniqueId, instanceId and incidenceId. 

How about if we keep uid/unique id names reserved for the ICS event uid since it's so much steering the assumptions on the APIs. InstanceId already used in kf5-calendarcore so could be clearer if it meant only that, leaving incidenceId for the nemo-calendar to use.  This latter one less much steering thinking as it's not in the spec, but since we can avoid the naming collision could do that.

If using names like this, would then need some additional adjustments to the qml api roles, and more code churn, but then again the usage is often needing adaption to the recurrenceId removal anyways?

src/calendarutils.cpp
dcaliste: Ok, incidenceId is the id returned by incidence->uid(). It is only required in the dissociate case, because the dissociation and the save are separated at the moment. The dissociation is done in one place, creates a Data::Event structure, pass it to the manager which pass it back with the save command, which in turn creates a new KCalendarCore::Event from the Data::Event.

dcaliste:
Ok, I can understand the confusion. So I've rework this PR and the associated one to follow these rules:
- `instanceId` is the unique identifier that span over notebooks, incidences and exceptions. Currently, it is implemented as `incidence->instanceIdentifier()` but it will also include the notebookUid in the future.
- `incidenceUid` is the uid as returned by `incidence->uid()`. It is a bit useless, but still mandatory because of the dissociation code. Currently the dissociation based on `instanceId` fetch the parent and create an exception incidence. This incidence has the right uid. But it is converted to Data::Event to be passed back to the manager, and the incidence->uid() is then lost because `event.instanceId` is forged from `exception->instanceIdentifier()`. Then in save, the Data::Event is passed and the routine detect that it should be an exception, but don't know the proper parent uid if not saved.
- `recurrenceId` is the exception tag, if any.
- `calendarUid` is the notebook uid.

Everywhere in the code, I've followed these naming conventions and changed public API and class members to reflect these.

Do you feel it better ?

pvuorela:
Looked a lot easier to grasp now, thanks. Couple comments for apis with instanceid + recurrence id but no event uid.

lightweight/common/eventdata.cpp
```diff
@@ -4,7 +4,7 @@ QDBusArgument &operator<<(QDBusArgument &argument, const EventData &eventData)
 {
     argument.beginStructure();
     argument << eventData.calendarUid
-             << eventData.uniqueId
+             << eventData.instanceId
```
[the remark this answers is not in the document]
dcaliste: Good catch. I would vote to remove it completely from the lightweight interface.
`grep -lr recurrenceId /usr/lib64/qt5/qml/` is not returning any match and `grep -lr recurrenceId /usr/share` returns a single match in the DBus calendar interface, which is expected. There is another match in /usr/share/jolla-calendar/pages/EventEditPage.qml in a test to know if the event is an exception.

So as far as I know with this information, it is safe to remove it from the lightweight interface. As you mentioned, one could add the incidenceUid instead. But removing the recurrenceId somehow hides an implementation detail that is not used. If we want to expose the parent of an exception at one moment, it would be better to send the instanceId of the parent instead.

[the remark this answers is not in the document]
pvuorela: Removal is fine here too. Either way.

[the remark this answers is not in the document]
dcaliste: I've removed it, it makes things simpler. We can still add it later if needed, like we did for the cancel status for instance.

src/calendarworker.cpp
```diff
@@ -319,11 +340,11 @@ void CalendarWorker::saveEvent(const CalendarData::Event &eventData, bool update
...
 }
 
-CalendarData::Event CalendarWorker::dissociateSingleOccurrence(const QString &uid, const QDateTime &recurrenceId)
+CalendarData::Event CalendarWorker::dissociateSingleOccurrence(const QString &instanceId, const QDateTime &recurrenceId)
 {
-    KCalendarCore::Event::Ptr event = mCalendar->event(uid);
-    if (!event) {
-        qWarning("Event to create occurrence replacement for not found");
+    KCalendarCore::Incidence::Ptr event = mCalendar->instance(instanceId);
+    if (!event || event->hasRecurrenceId()) {
```
pvuorela, line 346: Is this proper? So coming eventually from calendareventmodification. Should be possible to have an exception on recurring exception (i.e. repeats and has recurrence id due to replacing something earlier)? The old code allowed such so should this get the parent event if recurring?

src/calendarworker.cpp
```diff
@@ -221,18 +242,18 @@ bool CalendarWorker::sendResponse(const QString &uid, const QDateTime &recurrenc
...
 
     return sent;
 }
 
-QString CalendarWorker::convertEventToICalendar(const QString &uid, const QString &prodId) const
+QString CalendarWorker::convertEventToICalendar(const QString &instanceId, const QString &prodId) const
 {
     // NOTE: not fetching eventInstances() with different recurrenceId
-    KCalendarCore::Event::Ptr event = mCalendar->event(uid);
+    KCalendarCore::Incidence::Ptr event = mCalendar->instance(instanceId);
```
pvuorela, line 254: Earlier fetched the main event by just using uid. Should that be taken into account, converting instance with recurrence id? Should fetch the parent if recurring?

src/calendarmanager.h
pvuorela: Or well, this goes eventually to calendarworker.cpp and to ExtendedCalendar::dissociateSingleOccurrence() where there's a datetime parameter, but it's called just "datetime" instead of recurrenceId.

So maybe we should be calling the parameter on the similar methods as "datetime" or something like that?

src/calendarevent.h
```diff
@@ -62,7 +62,7 @@ class CalendarEvent : public QObject
     Q_PROPERTY(CalendarEvent::Days recurWeeklyDays READ recurWeeklyDays NOTIFY recurWeeklyDaysChanged)
     Q_PROPERTY(int reminder READ reminder NOTIFY reminderChanged)
     Q_PROPERTY(QDateTime reminderDateTime READ reminderDateTime NOTIFY reminderDateTimeChanged)
-    Q_PROPERTY(QString uniqueId READ uniqueId NOTIFY uniqueIdChanged)
+    Q_PROPERTY(QString instanceId READ instanceId NOTIFY instanceIdChanged)
```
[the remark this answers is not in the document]
pvuorela: Sure. Guess that's fine.

pvuorela (commented):
Some small comments still. 

One thing I'm still a bit thinking is how it goes when events were referred just by uid to e.g. load al the related data and then marked as such. Now there's some code for loading the main event instead of exception if instanceId is such, but then the loaded is marked by instanceId. Though it used to be a bit like that already. Maybe it's all good, but didn't wrap my head around it totally yet.

If you make changes, could rebase.

src/calendarevent.cpp
```diff
@@ -248,17 +241,17 @@ void CalendarStoredEvent::notebookColorChanged(QString notebookUid)
 
 void CalendarStoredEvent::eventUidChanged(QString oldUid, QString newUid)
 {
-    if (mData->uniqueId == oldUid) {
-        mData->uniqueId = newUid;
-        emit uniqueIdChanged();
+    if (mData->instanceId == oldUid) {
```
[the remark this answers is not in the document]
dcaliste: It's legacy code. As you notice, it's currently dead code because the manager is never emitting the signal.

I'm not touching it (besides member naming change) for the moment.

src/calendarworker.cpp
dcaliste: There is no support at the moment for recurring exceptions in KCalendarCore, as far as I know.

At the moment in the UI, there is no way to make an exception out of an exception. In the sense that a CalendarEventModification that call ::dissociateSingleOccurrence() can be called only when modifying a recurring parent and selecting "modify this occurrence only".

Previously, the code allowed was adding an exception to a recurring parent (only the uid was fetch, without recurrenceId). So here the code is still doing the same, checking that the instanceId was not belonging to an exception.

That being said, to get the exact same behaviour (the caller is an exception, but only pass the uid), I can change for the following:
```
KCalendarCore::Incidence::Ptr event = mCalendar->instance(instanceId);
if (event && event->hasRecurrenceId()) {
    event = mCalendar->incidence(event->uid());
}
if (!event) {
    qWarning(...
```

I thought it was a bit verbose for a case that is not possible from UI design neither from internal KCalendarCore implementation.

src/calendarworker.cpp
dcaliste: I'm happy you raise this point. I noticed it when making the changes.

In the previous code, if you export an exception from the UI, it will actually create a VEVENT for the parent only. There is no appending of the exceptions or what ever. To keep it simple, I decided to change to actually export the exception only in that case.

It's maybe not the best idea I had, and I would understand if you think it's better not to change any behaviour with this commit and possibly address it in a later dedicated PR.

src/calendarmanager.h
dcaliste: Very good suggestion, renaming the argument to `datetime` definitely reduce the confusion. Because this is _not_ a `recurrenceId` as used everywhere else to identify an exception. Internally, the recurrenceId will be set from `datetime` but it may be completely different with a different implementation.

I've changed the parameter name in the manager and in the worker. Thank you for the suggestion.

dcaliste:
> One thing I'm still a bit thinking is how it goes when events were referred just by uid to e.g. load al the related data and then marked as such. Now there's some code for loading the main event instead of exception if instanceId is such, but then the loaded is marked by instanceId. Though it used to be a bit like that already.

Exactly, this is all the case already. What is currently (before this PR) happening is the following:
- mKCal expose a `loadIncidenceByInstance(instanceId)` that loads into memory the fully series containing `instanceId`.
- the worker is calling the former, passing `instanceId` and send the full series to the manager thanks to a loop on all the events in memory (there is a filtering done with the `mSentEvents` hash to avoid sending twice the same instance between calls).
- the manager is storing each instance of the series in a hash table indexed by instanceId.
- all current routines in the manager using a (uid, recId) tuple are creating the corresponding instanceId and are thus accessing the right CalendarData::Event.

This PR is only changing the last step, using directly the instanceId in its API instead of (uid, recId) tuples. As far as I know, there is no code that loads a parent incidence (alone) when passed an (exception) instanceId and flagging it with the exception instanceId. In the worker, the fully series is always loaded and then the CalendarData::Events are sent to the manager, indexed using their own instanceId.

src/calendarworker.cpp
pvuorela: At least we've disabled recurrence on the jolla-calendar side for exceptions, but I'm not sure anymore was that to keep it just simpler or were there deeper problems. But if we do e.g. that "modify this and later" to change time of a meeting it should be done by modifying the exception and needs repetition?

dcaliste: The this and later modifications, still didn't make in. But it's a good argument.

The implementation of the this and later does not require to dissociate from an exception. At least not the way it is implemented in KCalendarCore (it may be broken though, don't know). The exception is always dissociated from the parent. Actually, dissociation means internally "provide a recurrenceId that match one of the occurrence of the _parent_" and not one of the occurrence of the current this and later exception.

So let say you have a recurring parent P, recurring every day at noon.
- Then at one point in time, you decide that it should start recurring every day at 1pm. It means to create a non recurring clone of P and `setRecurrenceId(noon on day D1)`, `setThisAndFuture(true)` and `setDtStart(1pm on day D1)`.
- Later, you decide to move it every day at 11am. It means to create a non recurring clone of P and `setRecurrenceId(noon on day D2)`, `setThisAndFuture(true)` and `setDtStart(11am on day D2)`.

So the call to `dissociateSingleOccurrence()` must always be called with the `instanceId` of the parent.

dcaliste:
Rebased on latest master.

pvuorela (approved):
I think this is good to go in. Thanks for your efforts!

PR #61, opened 2023-06-29
[nemo-qml-plugin-calendar] Clone existing parent when saving an excep…

dcaliste:
…tion.

When moving the dissociation code out of the
save function, it had the drawback that only
properties copied in the CalendarData::Event
structure are preserved in the exception.

This may creates some issues with the sync
plugins that may use properties that are not
part of CalendarData::Event structure.

Ensure that such properties are kept when
creating an exception by cloning the parent
before updating it with the exception
modifications.

@pvuorela, this is correcting an issue introduced in 4e0b556a, when moving the dissociation code out of the save routine. This move simplified the creation of a CalendarEventModification but had the drawback that some properties of the parent (those not duplicated in the CalendarData::Event structure) were lost in the process. This PR should solve this. It would also make sailfishos/buteo-sync-plugin-caldav#20 obsolete.

### DIFF
--- a/src/calendarworker.cpp
+++ b/src/calendarworker.cpp
@@ -284,48 +284,58 @@ void CalendarWorker::saveEvent(const CalendarData::Event &eventData, bool update
     KCalendarCore::Event::Ptr event;
     if (!eventData.instanceId.isEmpty()) {
         event = mCalendar->instance(eventData.instanceId).staticCast<KCalendarCore::Event>();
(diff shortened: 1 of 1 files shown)


dcaliste:
Going back to my calendar modifications, @pvuorela, may you have a look on this PR when you have the time to, please ? Thank you in advance.

dcaliste:
Thanks, no problem. No hurry.

pvuorela (approved):
Seems sensible.

PR #62, opened 2023-11-08
[nemo-qml-plugin-email] Avoid cache issue on reimportation.

dcaliste:
Import an ICS file once and delete the entry.
On reimportation of the same ICS file, but
to a different notebook, the old notebook
is still in use.

This is a backport of the implementation done
in cfd2ffec but not assuming that the events
are stored by instanceIdentifier.

@pvuorela, in case there will be another hot fix for the 4.5.0 branch. Here is a partial backport to avoid having cache issues when importing an event to a calendar, deciding differently, delete it and reimport it to a different calendar. The idea here is the same than the one in cfd2ffec : send notification to every objects that may have changed because of data loading. Before, this action was done in a bit convoluted way by listing the old CalendarData::Event with a matching object before sending the change notifications. The problem being that mEvents holding the old CalendarData::Event structures may be flushed, while the mEventObjects is never flushed and contains their own copy of CalendarData::Event.

### DIFF
--- a/src/calendarmanager.cpp
+++ b/src/calendarmanager.cpp
@@ -739,13 +739,6 @@ void CalendarManager::dataLoadedSlot(const QList<CalendarData::Range> &ranges,
                                      const QHash<QDate, QStringList> &dailyOccurrences,
                                      bool reset)
 {
-    QList<CalendarData::Event> oldEvents;
-    foreach (const QString &uid, mEventObjects.keys()) {
-        // just add all matching uid, change signal emission will match recurrence ids
-        if (events.contains(uid))
-            oldEvents.append(mEvents.values(uid));
-    }
-
     if (reset) {
         mEvents.clear();
         mEventOccurrences.clear();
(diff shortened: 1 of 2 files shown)


pvuorela:
Per se the change is quite similar to the earlier change, but it does make me a bit uneasy that this is anyway new code, I don't have 4.5.0 setup to test with right now, the 4.5.0 getting old already and the fixed bug shouldn't be that common to trigger.

Thinking of leaving this open for now and reconsider if such a hotfix is progressing.

dcaliste:
> I don't have 4.5.0 setup to test with right now

Yeh, I add to bring back my calendar stack to 4.5.0 myself also to reproduce the problem and understand the cache issue. Actually, I didn't completely bring back the system to 4.5.0, but instead decompress the rpm of `mkcal`, `nemo-qml-plugin-calendar` and `jolla-calendar` into a temporary directory, and then ran `jolla-calendar` with various environment variables to use code from this temporary directory, something like:
```
cd tmp
LD_LIBRARY_PATH=$PWD/lib64 SQLITESTORAGEDB=db QML2_IMPORT_PATH=$PWD/lib64/qt5/qml APP_PATH=$PWD/share/jolla-calendar/ bin/jolla-calendar
```
(it just required to slightly modified the `jolla-calendar` executable to set the QML root from the `APP_PATH` variable instead of the hardcoded `DEPLOYMENT_PATH` macro.)

All this being said, no problem of course to delay this to a hypothetic 4.5.0 hot fix. Anyway, I compiled this patch for the user who contacted me and he's happy with it.

pvuorela:
Don't think we'll do another 4.5.0 by this point so I'll close this one.

PR #71, opened 2026-03-13
Avoid discard warning

pvuorela:
Let's finally avoid getting this "No transactionId, discarding results" warning repeatedly in journal.

cc @dcaliste


dcaliste:
It looks correct to me. Nice bonus for new connect syntax  and other cleanups.

direc85 (approved):
One more meaningless warning taken care of!

PR #73, opened 2026-04-16
feat: port to Qt6

PureTryOut:
Requires Qt6 support in mkcal which isn't merged yet, https://github.com/sailfishos/mkcal/pull/75

### DIFF
--- a/lightweight/calendareventsmodel/calendareventsmodel.cpp
+++ b/lightweight/calendareventsmodel/calendareventsmodel.cpp
@@ -223,13 +223,13 @@ QVariant CalendarEventsModel::data(const QModelIndex &index, int role) const
         return eventData.description;
     case StartTimeRole:
         if (eventData.allDay) {
-            return QDateTime(QDate::fromString(eventData.startTime, Qt::ISODate));
+            return QDateTime(QDate::fromString(eventData.startTime, Qt::ISODate), QTime());
         } else {
             return QDateTime::fromString(eventData.startTime, Qt::ISODate);
         }
     case EndTimeRole:
         if (eventData.allDay) {
-            return QDateTime(QDate::fromString(eventData.endTime, Qt::ISODate));
+            return QDateTime(QDate::fromString(eventData.endTime, Qt::ISODate), QTime());
         } else {
             return QDateTime::fromString(eventData.endTime, Qt::ISODate);
         }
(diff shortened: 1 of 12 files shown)


tests/tst_calendarevent/tst_calendarevent.cpp
```diff
@@ -241,8 +241,8 @@ void tst_CalendarEvent::testSave()
 
     // mKCal DB stores times as seconds, losing millisecond accuracy.
     // Compare dates with QDateTime::toTime_t() instead of QDateTime::toMSecsSinceEpoch()
-    QCOMPARE(eventB->endTime().toTime_t(), endTime.toTime_t());
-    QCOMPARE(eventB->startTime().toTime_t(), startTime.toTime_t());
+    QCOMPARE(eventB->endTime().toSecsSinceEpoch(), endTime.toSecsSinceEpoch());
```
pvuorela, line 245: toSecsSinceEpoch introduced in qt 5.8. We still need support for 5.6. Older toMSecsSinceEpoch() is probably fine here even without the usual "/ 1000".

lightweight/calendareventsmodel/calendareventsmodel.cpp
```diff
@@ -223,13 +223,21 @@ QVariant CalendarEventsModel::data(const QModelIndex &index, int role) const
         return eventData.description;
     case StartTimeRole:
         if (eventData.allDay) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+            return QDateTime(QDate::fromString(eventData.startTime, Qt::ISODate), QTime());
```
pvuorela, line 226: Suppose the explicit QTime() should be fine even on older Qt versions.

src/calendarmanager.cpp
```diff
@@ -757,7 +757,11 @@ void CalendarManager::dataLoadedSlot(const QList<CalendarData::Range> &ranges,
 
     m_loadedRanges = addRanges(m_loadedRanges, ranges);
     m_loadedQueries.append(instanceList);
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
pvuorela: QHash::insert(QHash) looks like 5.15 addition.

pvuorela (approved):
Don't have qt6 setup for this but generally looks about right and qt5 not that much changed. Thanks!

